PR #11, opened 2023-01-12
Added pytest-cov to req-dev.txt to fix pytest

zacharybinger:
pytest-cov is required when running tests due to the use of --cov arguments in the pytest.ini file


kurbansitterley:
Incorporated into #14.

PR #12, opened 2023-01-12
Fix issues in GitHub Actions workflow configuration

lbianchi-lbl:


lbianchi-lbl:
@kurbansitterley I've fixed the error in the CI configuration that prevented the workflow to run, plus a couple more errors (including applying the same changes proposed in #11). The test suite is now running, and to a first look it seems that the failures are "genuine" (i.e. not due to configuration/environment errors).

kurbansitterley:
@lbianchi-lbl I believe I fixed these to some extent with #14 and #15 (and an errant push to main), but test with noneditable pkg are still failing, so I took that out of the GHA config in #15. Any idea why noneditable might be failing?

lbianchi-lbl:
> @lbianchi-lbl I believe I fixed these to some extent with #14 and #15 (and an errant push to main), but test with noneditable pkg are still failing, so I took that out of the GHA config in #15. Any idea why noneditable might be failing?

I believe the reason why it fails is that options for the `pytest-coverage` pytest plugin are specified in `pytest.ini`:
https://github.com/watertap-org/watertap-seto/blob/77b555a87fb6829bfece7a9e414b44910834e2d4/pytest.ini#L3

However, that plugin is only installed in the dev environments (it's specified in `requirements-dev.txt`), while the same `pytest.ini` is used both for dev and non-dev (non-editable) variants in CI. I'm not sure at what point that particular `pytest.ini` was added to the repo, but the `pytest-doverage` option only really makes sense within a dev environment.

Depending on your specific needs and/or preferences, you could consider removing that line from `pytest.ini` and add it in the CI job directly, although that will require a slight change in the CI configuration. I can try open a PR with this change at some point, but as I understand there's no pressing need since the CI seems to be passing now that the non-editable checks have been removed.

PR #13, opened 2023-01-23
Add SETOSystemCosting to aggregate system treatment and energy costs

kurbansitterley:
Improvements to costing approach for SETO-WaterTAP:

- Adds separate ``SETOSystemCosting`` class to aggregate system treatment and energy costs.
- Change ``factor_capital_annualization`` from being a fixed parameter to being calculated from an interest rate (``wacc`` - weighted average cost of capital) 
- Add helper functions for adding LCOW and LCOE calculations
- Separate methods for specific cost of electric energy and specific cost of thermal energy.
- Adds ``TreatmentCosting`` and ``EnergyCosting`` classes that inherit from ``SETOWaterTAPCosting`` (and are identical at this point).
- Adds USD2021 to units registry

### DIFF
--- a/src/watertap_contrib/seto/costing/seto_watertap_costing_package.py
+++ b/src/watertap_contrib/seto/costing/seto_watertap_costing_package.py
@@ -3,7 +3,7 @@ import pyomo.environ as pyo
 from idaes.core import declare_process_block_class
 from idaes.models.unit_models import Mixer
 
-from watertap.costing.watertap_costing_package import WaterTAPCostingData
+from watertap.costing.watertap_costing_package import WaterTAPCostingData, _DefinedFlowsDict
 from watertap.unit_models import (
     ReverseOsmosis0D,
     ReverseOsmosis1D,
@@ -19,7 +19,10 @@ from watertap.unit_models import (
     IonExchange0D,
     GAC,
 )
-
+from idaes.core.base.costing_base import (
+    FlowsheetCostingBlockData,
+    register_idaes_currency_units,
+)
 from watertap.costing.units.crystallizer import cost_crystallizer
 from watertap.costing.units.electrodialysis import cost_electrodialysis
 from watertap.costing.units.energy_recovery_device import cost_energy_recovery_device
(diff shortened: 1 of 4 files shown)


kurbansitterley:
Needed to do this to do RO+PV net metering demo

zacharybinger (approved):

src/watertap_contrib/seto/costing/seto_watertap_costing_package.py
```diff
@@ -63,7 +71,12 @@ class SETOWaterTAPCostingData(WaterTAPCostingData):
     def build_global_params(self):
         super().build_global_params()
 
-        self.base_currency = pyo.units.USD_2020
+        if "USD_2021" not in pyo.units._pint_registry:
+            pyo.units.load_definitions_from_strings(
+                ["USD_2021 = 500/708.0 * USD_CE500"]
+            )
+
+        self.base_currency = pyo.units.USD_2021
```
adam-a-a, line 79: I think this is now available in IDAES so maybe you can remove this.  

https://github.com/IDAES/idaes-pse/blob/a1e420575b1f0ffaaf15df2ff5882b8a9e39e260/idaes/core/base/tests/test_costing_base.py#L74so I wonder if this is still needed.

src/watertap_contrib/seto/costing/seto_watertap_costing_package.py
```diff
@@ -117,3 +130,353 @@ def build_process_costs(self):
...
+
+        self._registered_LCOWs = {}
+
+    def build_global_params(self):
+        # Register currency and conversion rates based on CE Index
+        register_idaes_currency_units()
+        if "USD_2021" not in pyo.units._pint_registry:
+            pyo.units.load_definitions_from_strings(
+                ["USD_2021 = 500/708.0 * USD_CE500"]
+            )
```
adam-a-a, line 166: Same as previous comment

src/watertap_contrib/seto/costing/seto_watertap_costing_package.py
kurbansitterley: We are using IDAES 2.0.0b2, I don't see USD2021 in the units: https://github.com/IDAES/idaes-pse/blob/0bdd637988ede81c1142096b480964759c98467e/idaes/core/base/costing_base.py#L84

adam-a-a (approved):
LGTM- we can come back and refine as needed later if needed. Also will make an issue corresponding to version of IDAES and use of latest currency year.

PR #16, opened 2023-01-23
Initial demonstration of RO+PV net metering

kurbansitterley:
Initial demonstration of net metering for SETO-WaterTAP using RO coupled to the PySAM PV model. Includes:
- WaterTAP-SETO analysis directory creation
- Flowsheet buid and run in ``ro_pv.py``
- Results display function in ``util.py``
- PySAM config files


zacharybinger (approved):

src/watertap_contrib/seto/analysis/net_metering/pysam_data/utilityrate5.json
```diff
@@ -0,0 +1,4 @@
+{
+	"en_electricity_rates" : 0,
+	"degradation" : [ 0 ]
```
adam-a-a, line 3: Why do we need this file?

src/watertap_contrib/seto/analysis/net_metering/ro_pv.py
```diff
@@ -0,0 +1,401 @@
...
+solver = get_solver()
+
+absolute_path = os.path.dirname(__file__)
+print(absolute_path)
+
+tech_config_file = "/pysam_data/pvsamv1.json"
+tech_config_file = absolute_path + tech_config_file
+grid_config_file = "/pysam_data/grid.json"
+grid_config_file = absolute_path + grid_config_file
+rate_config_file = "/pysam_data/utilityrate5.json"
```
adam-a-a, line 59: I suppose that file I questioned earlier would be modified to match some desired electricity rate.

src/watertap_contrib/seto/analysis/net_metering/pysam_data/utilityrate5.json
kurbansitterley: I don't know exactly. This is a question for @Matthew-Boyd.

src/watertap_contrib/seto/analysis/net_metering/ro_pv.py
```diff
@@ -0,0 +1,401 @@
...
+cash_config_file = "/pysam_data/singleowner.json"
+cash_config_file = absolute_path + cash_config_file
+weather_file = "/pysam_data/phoenix_az_33.450495_-111.983688_psmv3_60_tmy.csv"
+weather_file = absolute_path + weather_file
+
+
+def build_ro_pv():
+    m = ConcreteModel()
+    m.fs = FlowsheetBlock(dynamic=False)
+    m.db = SETODatabase()
```
adam-a-a, line 70: Since I am starting with this PR, I'm unaware what the purpose of `SETODatabase` is. I am guessing you have some cost data/assumptions there but haven't looked into it. Can you tell me a bit about it?

src/watertap_contrib/seto/analysis/net_metering/ro_pv.py
```diff
@@ -0,0 +1,401 @@
...
+    m.pysam = PySAMWaterTAP(
```
adam-a-a, line 71: What does this class do exactly, and why is it assigned as `m.pysam`?

src/watertap_contrib/seto/analysis/net_metering/ro_pv.py
```diff
@@ -0,0 +1,401 @@
...
+        rate_config_file=rate_config_file,
+        cash_config_file=cash_config_file,
+        weather_file=weather_file,
+    )
+    m.fs.properties = NaClParameterBlock()
+
+    treatment = m.fs.treatment = Block()
+    energy = m.fs.energy = Block()
+
+    energy.pv = PhotovoltaicZO(property_package=m.fs.properties, database=m.db)
```
adam-a-a, line 84: Why would PV be described by the NaCl property package?

adam-a-a: Is it just filler/dummy prop model since PV wouldn't be related to a prop package?

adam-a-a:
@kurbansitterley I still need to go through this fully, but I dropped some initial naive comments/questions to get started.

src/watertap_contrib/seto/analysis/net_metering/ro_pv.py
kurbansitterley: At this point it is identical to the ``Database`` class in WaterTAP. I created a separate class initially thinking we would be dealing with different types of data (e.g., location, weather) in SETO, and might need some special functionality at some point. But using the WaterTAP ``Database`` would work in this flowsheet.

The PV model is a ZO model, so requires a database, but the ``PhotovoltaicZO`` model could be changed to a more straightforward structure (e.g. just being a ``Block``). I am open to suggestions.

src/watertap_contrib/seto/analysis/net_metering/ro_pv.py
kurbansitterley: It is just a vessel through which to run PySAM models right now. No special reason it is called ``m.pysam`` and having it live on the model block makes it easier to find (i.e. see ``_get_pysam`` method in ``SETOSystemCosting`` class, needed for adding LCOE)

src/watertap_contrib/seto/analysis/net_metering/ro_pv.py
kurbansitterley: > Is it just filler/dummy prop model since PV wouldn't be related to a prop package?

Yes. Just a passthrough dummy ZO model. Per our last meeting, there is a better way to do it, I just haven't done it yet for this net metering demo.

src/watertap_contrib/seto/analysis/net_metering/pysam_data/utilityrate5.json
Matthew-Boyd: This is one of the four files that contain PySAM model parameters, each for their respective submodels. Organizing the parameters in files this way is also how they are exported from the SAM UI and is thus the easiest way to transfer parameters to PySAM from a SAM project.

src/watertap_contrib/seto/analysis/net_metering/PV_RO.py
```diff
@@ -0,0 +1,457 @@
...
+        rate_config_file=rate_config_file,
+        cash_config_file=cash_config_file,
+        weather_file=weather_file,
+    )
+    m.fs.properties = NaClParameterBlock()
+
+    treatment = m.fs.treatment = Block()
+    energy = m.fs.energy = Block()
+
+    energy.pv = PhotovoltaicZO(property_package=m.fs.properties, database=m.db)
```
adam-a-a, line 93: I think we should change this into a regular unit model (as opposed to ZO model) and just exclude the property package related config options. However, for the sake of moving things along, we can reserve such changes for a subsequent PR.

src/watertap_contrib/seto/analysis/net_metering/PV_RO.py
```diff
@@ -0,0 +1,457 @@
+import os
```
adam-a-a, line 1: We really should add a test file for this. Something is better than nothing.

adam-a-a (commented):
I'm sure there are a bunch of things that we might need to refine in the near-future, but we can reserve all of that for a subsequent PR as needed. However, I think we should at least add a test file for PVRO. It would help serve as a first checkpoint to ensure that we don't break things with subsequent changes (or to discover things that are breakable that should be fixed).

src/watertap_contrib/seto/analysis/net_metering/PV_RO.py
kurbansitterley: I was thinking about this yesterday and agree. @zacharybinger can you provide a simple test file for the RO+PV flowsheet?

src/watertap_contrib/seto/analysis/net_metering/PV_RO.py
kurbansitterley: Yes agreed- this implementation was meant to be temporary but caught legs and here we are. I will address this and might as well do it before we get the test file in so we don't have to redo that as well.

src/watertap_contrib/seto/analysis/net_metering/PV_RO.py
```diff
@@ -0,0 +1,457 @@
...
+    m.fs.energy.costing.factor_maintenance_labor_chemical.fix(0)
+
+
+def solve(m, solver=None, tee=False, check_termination=True):
+    if solver is None:
+        solver = get_solver()
+    results = solver.solve(m, tee=tee)
+    size_pv(m)
+    fix_pysam_costing(m)
+    results = solver.solve(m, tee=tee)
```
Matthew-Boyd, line 419: Have you done any comparisons with the SAM UI to ensure your dependent parameters for costing and the DC-to-AC ratio update as expected in this Python model, and that the results agree? If not, I would highly recommend.

Matthew-Boyd (dismissed):
I don't see any glaring issues with how you're utilizing the PV model via PySAM. I strongly recommend comparing the functionality of your parameter relations to those in the UI, however, if you haven't already.

Also, I see that you're zeroing the hourly energy when less than 0, on line 102 in pysam_watertap.py 
self.hourly_energy = [x if x > 0 else 0 for x in self.hourly_energy]

Note that negative values indicate a power draw from the grid, so you will want to account for that if you're not somewhere else.

kurbansitterley:
Tests are failing here but passing locally for me. I believe it is related to [WaterTAP Issue 929](https://github.com/watertap-org/watertap/issues/929)

kurbansitterley:
LT-MED tests are failing, not PV/RO. I think it is a solver issue. I am merging this so we can get relevant changes in and move on with more analysis. We can address test failings later if they persist.

PR #18, opened 2023-01-24
Add initialize method to LT-MED model

kurbansitterley:
Proposed changes:

- Add ``initialize_build`` method to LT-MED model
- Change ``feed_props`` to having defined state
- Add initialization test to test file
- Change ``recovery_ratio`` name to ``recovery_vol_phase`` indexed to time and ``"Liq"``
- Change ``specific_thermal_energy_consumption`` to ``specific_energy_consumption_thermal``
- Make necessary changes to results tests

### DIFF
--- a/src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
+++ b/src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
@@ -11,6 +11,8 @@
 #
 ###############################################################################
 
+from copy import deepcopy
+
 # Import Pyomo libraries
 from pyomo.environ import (
     Set,
@@ -18,6 +20,7 @@ from pyomo.environ import (
     Param,
     Suffix,
     Constraint,
+    check_optimal_termination,
     units as pyunits,
 )
 from pyomo.common.config import ConfigBlock, ConfigValue, In
@@ -28,7 +31,9 @@ from idaes.core import (
     UnitModelBlockData,
     useDefault,
 )
+from idaes.core.solvers.get_solver import get_solver
 from idaes.core.util.config import is_physical_parameter_block
+from idaes.core.util.exceptions import ConfigurationError, InitializationError
 import idaes.core.util.scaling as iscale
 import idaes.logger as idaeslog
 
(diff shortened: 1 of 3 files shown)


src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
```diff
@@ -759,21 +770,146 @@ def eq_cool_vol_flow(b):
...
+                if state_dict[k].is_indexed():
+                    state_args[k] = {}
+                    for m in state_dict[k].keys():
+                        state_args[k][m] = state_dict[k][m].value
+                else:
+                    state_args[k] = state_dict[k].value
+
+        state_args_dist = deepcopy(state_args)
+        for p, j in blk.distillate_props.phase_component_set:
+            if j == "TDS":
```
adam-a-a, line 828: Should we generalize this to check that component is Solute() rather than a hard-coded "TDS"?

kurbansitterley: Actually I think the solute (and only solute) for this surrogate _has_ to be ``TDS``, so maybe we should put a check for that somewhere? This is a better question for @Zhuoran29

adam-a-a: Right, at this point, the surrogate applies to seawater, so TDS does make sense. I just thought it could be beneficial to keep it generic now in case there were ever an update to the model (even though I doubt it).

src/watertap_contrib/seto/unit_models/surrogate/tests/test_lt_med_surrogate.py
```diff
@@ -59,15 +62,17 @@ def LT_MED_frame(self):
             (sys_capacity / recovery_ratio), to_units=pyunits.m**3 / pyunits.s
         )
 
-        feed.conc_mass_phase_comp["Liq", "TDS"].fix(feed_salinity)
-        feed.flow_vol_phase["Liq"].fix(feed_flow)
+        # feed.conc_mass_phase_comp["Liq", "TDS"].fix(feed_salinity)
+        # feed.flow_vol_phase["Liq"].fix(feed_flow)
```
adam-a-a, line 66:
```suggestion
```

adam-a-a (approved):
LGTM - I won't let my comment about changing `TDS` to a check for `Solute()` slow this down.

PR #20, opened 2023-01-31
Add solar trough net-metering surrogate model

Matthew-Boyd:
Add a concentrating solar power (CSP) trough net-metering surrogate model that can be incorporated into flowsheets.

- Training data from simulations of the industrial process heat (IPH) trough model in NREL's SAM software is included.
- A script to run SAM via PySAM to generate training data is included. This file utilizes multiprocessing to speed up the simulations by running them in parallel. A thermal storage cost relation is added that is missing in PySAM.
- Adds a script for training a surrogate model, via Pysmo, from the included training data, and includes a trained surrogate model. This model predicts the annual heat energy and annual electrical load as a function of plant heat capacity and rated hours of storage.
- Adds a pyomo cost model for the trough.
- Contains a tests file that verifies the above functionality.


src/watertap_contrib/seto/costing/solar/trough.py
```diff
@@ -0,0 +1,161 @@
...
+
+
+def build_trough_cost_param_block(blk):
+
+    costing = blk.parent_block()
+
+    blk.base_storage_hours = pyo.Var(
+        initialize=6,
+        units=pyo.units.hour,
+        bounds=(6, 6),
```
adam-a-a, line 17: I would remove these bounds, or if the intention is to (almost) never change this, you could either (1) make this a mutable parameter or (2) keep this variable declaration and fix it to the value of 6 right after.

adam-a-a: IMHO - it makes more sense to just make this a mutable Parameter.

src/watertap_contrib/seto/costing/solar/trough.py
```diff
@@ -0,0 +1,161 @@
...
+    )
+
+    blk.cost_per_storage_capital = pyo.Var(
+        initialize=62,
+        units=costing.base_currency / pyo.units.kWh,
+        bounds=(0, None),
+        doc="Cost per kWh (thermal) for the trough plant thermal storage",
+    )
+
+    blk.contingency_frac_direct_cost = pyo.Var(
```
adam-a-a, line 35: I see how this makes sense at the unit level.

src/watertap_contrib/seto/costing/solar/trough.py
```diff
@@ -0,0 +1,161 @@
...
+        bounds=(0, 1),
+        doc="Fraction of direct costs applicable for sales tax",
+    )
+
+    blk.sales_tax_rate = pyo.Var(
+        initialize=0.05,
+        units=pyo.units.dimensionless,
+        bounds=(0, 1),
+        doc="Sales tax rate as applied to applicable capital costs",
+    )
```
adam-a-a, line 54: Would it make more sense to have these as flowsheet/system level variables?

src/watertap_contrib/seto/costing/solar/trough.py
```diff
@@ -0,0 +1,161 @@
...
+
+    blk.hours_storage = pyo.Var(
+        initialize=6,
+        units=pyo.units.hour,
+        bounds=(0, None),
+        doc="Rated hours of thermal storage",
+    )
+
+    blk.annual_generation = pyo.Var(
```
adam-a-a, line 109: I get that generation with respect to a parabolic trough implicitly means thermal generation. With that said, imagining future flowsheets where we might have, for example, parabolic troughs and PV, I think it would make sense to distinguish the type of generation rather than have the user insinuate the type of generation (as intuitive as it might be with the doc string and all).
```suggestion
    blk.annual_heat_generation = pyo.Var(
```

src/watertap_contrib/seto/costing/solar/trough.py
```diff
@@ -0,0 +1,161 @@
...
+    )
+
+    blk.direct_cost_constraint = pyo.Constraint(
+        expr=blk.direct_cost
+        == (
+            blk.system_capacity
+            * (
+                trough_params.cost_per_capacity_capital
+                - trough_params.base_storage_hours
+                * trough_params.cost_per_storage_capital
```
adam-a-a, line 130: So do you subtract that amount of thermal storage that is already assumed to be aggregated within the system cost? Is that the idea? Just want to make sure I am understanding correctly.

src/watertap_contrib/seto/costing/solar/trough.py
```diff
@@ -0,0 +1,161 @@
...
+    blk.electricity_cost_constraint = pyo.Constraint(
+        expr=blk.electricity_cost
+        == trough_params.electricity_rate * blk.annual_electrical_load
+    )
+
+    # TODO: register the flows, e.g.,:
+    # blk.costing_package.cost_flow(blk.unit_model.electricity, "electricity")
+
+    # NOTE:
+    # What is costing.base_period, as borrowed from /costing/solar/photovoltaic.py ?
```
adam-a-a, line 161: `costing.base_period` is meant to represent the time horizon considered-- so in our case, it's `year`, represented as a unit.

src/watertap_contrib/seto/solar_models/surrogate/data/pysam_run.py
```diff
@@ -0,0 +1,326 @@
...
+import PySAM.Lcoefcr as lcoefcr
+
+
+def read_module_datafile(file_name):
+    with open(file_name, "r") as file:
+        data = json.load(file)
+    return data
+
+
+def load_config(modules, file_names=None, module_datas=None):
```
adam-a-a, line 21:
```suggestion
def load_config(modules, file_names=None, module_data=None):
```

src/watertap_contrib/seto/solar_models/surrogate/data/pysam_run.py
```diff
@@ -0,0 +1,326 @@
...
+    for i in range(len(modules)):
+        if file_names is not None:
+            data = read_module_datafile(file_names[i])
+        elif module_datas is not None:
```
adam-a-a, line 25:
```suggestion
        elif module_data is not None:
```

src/watertap_contrib/seto/solar_models/surrogate/data/pysam_run.py
```diff
@@ -0,0 +1,326 @@
...
+            data = module_datas[i]
```
adam-a-a, line 26:
```suggestion
            data = module_data[i]
```
This assumes that `modules` is the same length as `file_names` or `module_data`. Will that always be the case? 

Also, I am guessing that these can be a single string or a list of strings. Should we enforce types? Also, could you add some doc strings to explain args a bit more?

src/watertap_contrib/seto/costing/solar/trough.py
kurbansitterley: Yes I think so. How do you grab system level costing params in a unit level costing model?

src/watertap_contrib/seto/costing/solar/trough.py
kurbansitterley: Agreed. Not committing suggestion here because I am not sure where else it is used (yet).

src/watertap_contrib/seto/costing/solar/trough.py
kurbansitterley: Yes and I can't imagine why it would ever change (though we should still use the convention here).

src/watertap_contrib/seto/costing/solar/trough.py
Matthew-Boyd: Thanks, I'll change to a mutable Parameter.

src/watertap_contrib/seto/costing/solar/trough.py
Matthew-Boyd: Agreed.

src/watertap_contrib/seto/costing/solar/trough.py
Matthew-Boyd: Correct, I am manually adjusting the storage cost that is aggregated within the direct cost as PySAM is not automatically adjusting that cost after changes to storage size.

src/watertap_contrib/seto/costing/solar/trough.py
Matthew-Boyd: Thanks

src/watertap_contrib/seto/solar_models/surrogate/data/pysam_run.py
Matthew-Boyd: Changed.

src/watertap_contrib/seto/solar_models/surrogate/data/pysam_run.py
Matthew-Boyd: Changed.

src/watertap_contrib/seto/costing/solar/trough.py
Matthew-Boyd: Changed, along with the other reference.

kurbansitterley:
replaced by #30

PR #26, opened 2023-02-16
Add Electrocoagulation unit model and BasicWaterProperties property package

kurbansitterley:
Adds the Electrocoagulation model as presented by NMSU and corresponding costing model and test file.

Adds simple property package (basic_water_properties.py) to be used with EC model (and likely other NMSU models). State variables are volumetric flow and component concentration. That's it. It is slightly more robust than the ZO properties in WaterTAP, but not by much. Also adds corresponding test file.

Other minor changes to costing model. 

Remove SW prop pack test file.


src/watertap_contrib/seto/costing/units/electrocoagulation.py
```diff
@@ -0,0 +1,321 @@
...
+        ec.properties_in[0].flow_vol, to_units=pyo.units.Mgallons / pyo.units.day
+    )
+    flow_m3_yr = pyo.units.convert(
+        ec.properties_in[0].flow_vol, to_units=pyo.units.m**3 / pyo.units.year
+    )
+    blk.annual_sludge_flow = pyo.units.convert(
+        sum(
+            ec.properties_waste[0].flow_mass_comp[j] if j != "H2O" else 0
+            for j in ec.properties_waste[0].params.component_list
+        ),
```
adam-a-a, line 157: Why omit water?

src/watertap_contrib/seto/property_models/basic_water_properties.py
```diff
@@ -0,0 +1,401 @@
+"""
+This module contains a basic property package for simple water treatment models.
```
adam-a-a, line 2: Before I do anything here, can you update me on your plans for migration directly to WaterTAP, @kurbansitterley? I think I know how to fix the errors cropping up, but don't want to start commits if this will all be abandoned and put up as a separate PR on WaterTAP.

kurbansitterley: The plan is to essentially merge the EC ZO version on the WaterTAP branch with this model. However, that model is pretty much one constraint atm so it will be mostly this model. I have converted this model to use the ZO framework on the [ec_zo_model](https://github.com/watertap-org/watertap/tree/ec_zo_model) branch of WaterTAP, but it isn't quite in PR shape yet (but will be soon).

kurbansitterley:
@adam-a-a I am going to close this given the EC model will be going into main WaterTAP- but part of this PR is also the ``BasicWaterProperties`` property package I created as a dead simple property package to go along with these unit models we got from NMSU. I have directed @MuktaHardikar to use it for the chemical softening model, so will still come in via a separate PR.

PR #27, opened 2023-02-20
Add sales tax as global parameter

kurbansitterley:
Add sales tax as global parameter rather than a separate one for each costing model.


kurbansitterley:
LT-MED tests are failing here also.

PR #31, opened 2023-02-23
Add default property scaling factors to LT-MED test file

kurbansitterley:
Add default property scaling factors to LT-MED test file to pass GHA tests.


bknueven:
The failures on this PR are caused by https://github.com/watertap-org/watertap/pull/853. I'll try to track down what the specific issue is.

adam-a-a:
If this is tied to [#853](https://github.com/watertap-org/watertap/pull/853), then maybe there is something wrong with how pressure is addressed in the LT-MED model. Looking at the tests, I don't see pressure being fixed at the inlet. Furthermore, what that PR on watertap did was to add pressure effects into the enthalpy calculation, so if pressure is unspecified on a particular stateblock where it should be, that could cause a failure.

kurbansitterley:
When I start fixing temperature on the state blocks, the dof go down... I haven't tried pressure yet, though.

adam-a-a:
OK, that must be why LT-MED is broken. Pressure is now used to calculate enthalpy. I see that LT-MED uses enthalpy but doesn't specify pressure anywhere

Zhuoran29:
I'm trying to find which equations is causing the conflict, but I'm in a meeting now, I shall push the PR later so that you guys can review.

Zhuoran29:
> @kurbansitterley @bknueven @Zhuoran29 If this is tied to [#853](https://github.com/watertap-org/watertap/pull/853), then maybe there is something wrong with how pressure is addressed in the LT-MED model. Looking at the tests, I don't see pressure being fixed at the inlet. Furthermore, what that PR on watertap did was to add pressure effects into the enthalpy calculation, so if pressure is unspecified on a particular stateblock where it should be, that could cause a failure.

Yes the pressure is not fixed in LT-MED, but I think it's initialized with the default value in property packages if not fixed. And actually in MED-TVC model it can be initialized without specifying the pressure as well.

bknueven:
On my system I still see `test_mass_balance` failing, though the values are close.

I've adding isobaric constraints to the model, which should fix the issue stemming from https://github.com/watertap-org/watertap/pull/853

src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
```diff
@@ -275,6 +275,22 @@ def eq_brine_vol_flow(b):
...
+        def eq_feed_to_distillate_isobaric(b, t):
+            return b.feed_props[t].pressure == b.distillate_props[t].pressure
+
+        @self.Constraint(self.flowsheet().config.time, doc="Isobaric")
+        def eq_feed_to_brine_isobaric(b, t):
+            return b.feed_props[t].pressure == b.brine_props[t].pressure
+
+        @self.Constraint(self.flowsheet().config.time, doc="Isobaric")
+        def eq_feed_to_cooling_isobaric(b, t):
+            return b.feed_props[t].pressure == b.cooling_out_props[t].pressure
```
adam-a-a, line 292: I definitely see how this would be true (assuming equal pressure between feed, distillate and cooling source), but I can also see scenarios where this might not be the case. @Zhuoran29 - please confirm whether this is always the case for PSA's LT-MED model, or if you think there would be any reason to later do sensitivity analysis or just be able to have different pressures between these streams. If there is a reasonable chance that someone might want to do this for whatever reason, let's post an issue about it (i.e., LT-MED assumes isobaric conditions but can later add a config option that allows user to specify different pressures as needed at each of the stateblocks). Either way, I won't let it hold up this PR, but I'd like your feedback on it.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_lt_med_surrogate.py
```diff
@@ -65,12 +65,25 @@ def LT_MED_frame(self):
         feed.flow_mass_phase_comp["Liq", "TDS"].fix(feed_salinity * feed_flow)
         feed.flow_mass_phase_comp["Liq", "H2O"].fix(feed_dens * feed_flow)
         feed.temperature.fix(feed_temperature + 273.15)
+        feed.pressure.fix(101325)
```
adam-a-a, line 68: Yes!

src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
adam-a-a: Let's try to also consider this from a flowsheet perspective, where the inlet pressure to LT-MED could be different if receiving a stream at a different pressure from atmospheric, for example.

adam-a-a (approved):
LGTM - I just have one comment for everyone, but specifically to @Zhuoran29, concerning isobaric assumptions, which should generally suffice.

src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
Zhuoran29: @adam-a-a Yes, I confirm it is always the case in this configuration. The surrogate model was developed based on a very specific configuration (actually a bit different than the PSA pilot plant, and is believed better), and in that configuration the pressure are assumed the same across the flows.

So if we're going to develop a more detailed physical model in the future, we should make the pressure more flexible as you mentioned, but at least for the surrogate it shouldn't be changed, otherwise the surrogate equations won't be suitable.

Zhuoran29:
> LGTM - I just have one comment for everyone, but specifically to @Zhuoran29, concerning isobaric assumptions, which should generally suffice.

Please see response in the code session.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_lt_med_surrogate.py
```diff
@@ -65,12 +65,25 @@ def LT_MED_frame(self):
...
+        )
+        m.fs.water_prop.set_default_scaling(
+            "flow_mass_phase_comp", 1e3, index=("Liq", "TDS")
+        )
+        m.fs.steam_prop.set_default_scaling(
+            "flow_mass_phase_comp", 1e-2, index=("Liq", "H2O")
+        )
+        m.fs.steam_prop.set_default_scaling(
+            "flow_mass_phase_comp", 1, index=("Vap", "H2O")
+        )
```
Zhuoran29, line 86: Does the model still fail without resetting the scaling factors?

Zhuoran29:
All looks good to me! Just not sure if resetting the scaling factors is still needed.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_lt_med_surrogate.py
adam-a-a: When running a model we usually supply default scaling for the inlet mass flowrates. It might be that the case run without it, but good practice to this in general.

Zhuoran29: I see, it makes sense for mass flowrates to fit the actual applications.

PR #32, opened 2023-03-06
Modification of MED-TVC model based on PR#10 (which was closed)

Zhuoran29:
### Fixes/Resolves:
Replace and update PR#10.

- Modification of MED-TVC model based on the comments from  PR#10.
- Added initialization methods.
- Added costing package for MED-TVC (technically same with LT-MED cost model).

### DIFF
--- a/src/watertap_contrib/seto/costing/units/med_tvc_surrogate.py
+++ b/src/watertap_contrib/seto/costing/units/med_tvc_surrogate.py
@@ -1,0 +1,258 @@
+import pyomo.environ as pyo
+from watertap.costing.util import register_costing_parameter_block
+from watertap_contrib.seto.costing.util import (
+    make_capital_cost_var,
+    make_fixed_operating_cost_var,
+)
+
+# Costing equations from:
+# Kosmadakis G, Papapetrou M, Ortega-Delgado B, Cipollina A, Alarcón-Padilla D-C.
+# "Correlations for estimating the specific capital cost of multi-effect distillation plants
+#    considering the main design trends and operating conditions"
+# doi: 10.1016/j.desal.2018.09.011
+
+
+def build_med_tvc_surrogate_cost_param_block(blk):
+
+    costing = blk.parent_block()
+
+    blk.cost_fraction_evaporator = pyo.Var(
+        initialize=0.4,
+        units=pyo.units.dimensionless,
+        bounds=(0, None),
+        doc="Cost fraction of the evaporator",
+    )
+
+    blk.cost_fraction_maintenance = pyo.Var(
+        initialize=0.02,
+        units=pyo.units.dimensionless,
+        bounds=(0, None),
+        doc="Fraction of capital cost for maintenance",
+    )
+
+    blk.cost_fraction_insurance = pyo.Var(
+        initialize=0.005,
+        units=pyo.units.dimensionless,
+        bounds=(0, None),
+        doc="Fraction of capital cost for insurance",
+    )
+
+    blk.cost_storage_per_kwh = pyo.Var(
+        initialize=26,
+        units=costing.base_currency / pyo.units.kWh,
+        bounds=(0, None),
+        doc="Cost of thermal storage per kWh",
+    )
+
+    blk.cost_chemicals_per_vol_dist = pyo.Var(
+        initialize=0.04,
+        units=costing.base_currency / pyo.units.m**3,
+        bounds=(0, None),
+        doc="Cost of chemicals per m3 distillate",
+    )
+
+    blk.cost_labor_per_vol_dist = pyo.Var(
+        initialize=0.033,
+        units=costing.base_currency / pyo.units.m**3,
+        bounds=(0, None),
+        doc="Cost of labor per m3 distillate",
+    )
+
+    blk.cost_misc_per_vol_dist = pyo.Var(
+        initialize=0.033,
+        units=costing.base_currency / pyo.units.m**3,
+        bounds=(0, None),
+        doc="Cost of labor per m3 distillate",
+    )
+
+    blk.cost_disposal_per_vol_brine = pyo.Var(
+        initialize=0.02,
+        units=costing.base_currency / pyo.units.m**3,
+        bounds=(0, None),
+        doc="Cost of disposal per m3 brine",
+    )
+
+    blk.specific_energy_consumption_electric = pyo.Var(
+        initialize=1.5,
+        units=pyo.units.kWh / pyo.units.m**3,
+        bounds=(0, None),
+        doc="Specific electric energy consumption",
+    )
+
+    blk.hours_thermal_storage = pyo.Var(
+        initialize=0,
+        units=pyo.units.hr,
+        doc="Hours of thermal storage required",
+    )
+
+    blk.med_sys_A_coeff = pyo.Var(
+        initialize=6291,
+        units=pyo.units.dimensionless,
+        doc="LT-MED system specific capital A coeff",
+    )
+
+    blk.med_sys_B_coeff = pyo.Var(
+        initialize=-0.135,
+        units=pyo.units.dimensionless,
+        doc="LT-MED system specific capital B coeff",
+    )
+
+    blk.heat_exchanger_ref_area = pyo.Var(
+        initialize=302.01,
+        units=pyo.units.m**2 / (pyo.units.kg / pyo.units.s),
+        doc="Specific heat exchanger area from reference plant",
+    )
+
+    blk.heat_exchanger_exp = pyo.Var(
+        initialize=0.8,
+        units=pyo.units.dimensionless,
+        doc="Exponent for specific heat exchanger cost equation",
+    )
+
+    blk.fix_all_vars()
+
+
+@register_costing_parameter_block(
+    build_rule=build_med_tvc_surrogate_cost_param_block,
+    parameter_block_name="med_tvc_surrogate",
+)
+def cost_med_tvc_surrogate(blk):
+
+    med_tvc_params = blk.costing_package.med_tvc_surrogate
+    make_capital_cost_var(blk)
+    make_fixed_operating_cost_var(blk)
+
+    med_tvc = blk.unit_model
+    feed = med_tvc.feed_props[0]
+    dist = med_tvc.distillate_props[0]
+    brine = med_tvc.brine_props[0]
+    base_currency = blk.config.flowsheet_costing_block.base_currency
+
+    blk.total_system_cost = pyo.Var(
+        initialize=100,
+        units=base_currency,
+        doc="MED system cost",
+    )
+
+    blk.membrane_system_cost = pyo.Var(
+        initialize=100,
+        units=base_currency,
+        doc="Membrane system cost",
+    )
+
+    blk.evaporator_system_cost = pyo.Var(
+        initialize=100,
+        units=base_currency,
+        doc="Evaporator system cost",
+    )
+
+    blk.med_specific_cost = pyo.Var(
+        initialize=100,
+        units=pyo.units.USD_2018 / (pyo.units.m**3 / pyo.units.day),
+        doc="MED system cost per m3/day distillate",
+    )
+
+    blk.heat_exchanger_specific_area = pyo.Var(
+        initialize=100,
+        units=pyo.units.m**2 / (pyo.units.kg / pyo.units.s),
+        doc="Specific heat exchanger area",
+    )
+
+    blk.thermal_storage_capacity = pyo.Var(
+        initialize=5,
+        units=pyo.units.kWh,
+        doc="Thermal storage capacity",
+    )
+
+    blk.heat_exchanger_specific_area_constraint = pyo.Constraint(
+        expr=blk.heat_exchanger_specific_area
+        == pyo.units.convert(
+            med_tvc.specific_area / feed.dens_mass_phase["Liq"],
+            to_units=(pyo.units.m**2 * pyo.units.s) / pyo.units.kg,
+        )
+    )
+
+    blk.thermal_storage_capacity_constraint = pyo.Constraint(
+        expr=blk.thermal_storage_capacity
+        == med_tvc.thermal_power_requirement * med_tvc_params.hours_thermal_storage
+    )
+
+    blk.capacity = pyo.units.convert(
+        dist.flow_vol_phase["Liq"], to_units=pyo.units.m**3 / pyo.units.day
+    )
+
+    blk.annual_dist_production = pyo.units.convert(
+        dist.flow_vol_phase["Liq"], to_units=pyo.units.m**3 / pyo.units.year
+    )
+    blk.med_specific_cost_constraint = pyo.Constraint(
+        expr=blk.med_specific_cost
+        == (
+            med_tvc_params.med_sys_A_coeff
+            * blk.capacity**med_tvc_params.med_sys_B_coeff
+        )
+    )
+    blk.membrane_system_cost_constraint = pyo.Constraint(
+        expr=blk.membrane_system_cost
+        == blk.capacity
+        * (blk.med_specific_cost * (1 - med_tvc_params.cost_fraction_evaporator))
+    )
+
+    blk.evaporator_system_cost_constraint = pyo.Constraint(
+        expr=blk.evaporator_system_cost
+        == blk.capacity
+        * (
+            blk.med_specific_cost
+            * (
+                med_tvc_params.cost_fraction_evaporator
+                * (
+                    (
+                        blk.heat_exchanger_specific_area
+                        / med_tvc_params.heat_exchanger_ref_area
+                    )
+                    ** med_tvc_params.heat_exchanger_exp
+                )
+            )
+        )
+    )
+
+    blk.total_system_cost_constraint = pyo.Constraint(
+        expr=blk.total_system_cost
+        == blk.membrane_system_cost + blk.evaporator_system_cost
+    )
+
+    blk.capital_cost_constraint = pyo.Constraint(
+        expr=blk.capital_cost
+        == blk.total_system_cost
+        + med_tvc_params.cost_storage_per_kwh * blk.thermal_storage_capacity
+    )
+
+    blk.fixed_operating_cost_constraint = pyo.Constraint(
+        expr=blk.fixed_operating_cost
+        == blk.annual_dist_production
+        * (
+            med_tvc_params.cost_chemicals_per_vol_dist
+            + med_tvc_params.cost_labor_per_vol_dist
+            + med_tvc_params.cost_misc_per_vol_dist
+        )
+        + blk.total_system_cost
+        * (
+            med_tvc_params.cost_fraction_maintenance
+            + med_tvc_params.cost_fraction_insurance
+        )
+        + pyo.units.convert(
+            brine.flow_vol_phase["Liq"], to_units=pyo.units.m**3 / pyo.units.year
+        )
+        * med_tvc_params.cost_disposal_per_vol_brine
+    )
+
+    blk.heat_flow = pyo.Expression(
+        expr=med_tvc.specific_energy_consumption_thermal
+        * pyo.units.convert(blk.capacity, to_units=pyo.units.m**3 / pyo.units.hr)
+    )
+    blk.electricity_flow = pyo.Expression(
+        expr=med_tvc_params.specific_energy_consumption_electric
+        * pyo.units.convert(blk.capacity, to_units=pyo.units.m**3 / pyo.units.hr)
+    )
+
+    blk.costing_package.cost_flow(blk.heat_flow, "heat")
+    blk.costing_package.cost_flow(blk.electricity_flow, "electricity")
--- a/src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
+++ b/src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
@@ -1,0 +1,1394 @@
+###############################################################################
+# WaterTAP Copyright (c) 2021, The Regents of the University of California,
+# through Lawrence Berkeley National Laboratory, Oak Ridge National
+# Laboratory, National Renewable Energy Laboratory, and National Energy
+# Technology Laboratory (subject to receipt of any required approvals from
+# the U.S. Dept. of Energy). All rights reserved.
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#
+###############################################################################
+
+from copy import deepcopy
+
+# Import Pyomo libraries
+from pyomo.environ import (
+    Set,
+    Var,
+    Param,
+    Suffix,
+    Constraint,
+    check_optimal_termination,
+    units as pyunits,
+)
+from pyomo.common.config import ConfigBlock, ConfigValue, In
+
+# Import IDAES cores
+from idaes.core import (
+    FlowsheetBlock,
+    ControlVolume0DBlock,
+    declare_process_block_class,
+    UnitModelBlockData,
+    useDefault,
+)
+from idaes.core.util.config import is_physical_parameter_block
+import idaes.core.util.scaling as iscale
+from idaes.core.solvers import get_solver
+from idaes.core.util.exceptions import ConfigurationError, InitializationError
+import idaes.logger as idaeslog
+
+# Import Watertap packages
+from watertap.property_models.seawater_prop_pack import SeawaterParameterBlock
+from watertap.property_models.water_prop_pack import WaterParameterBlock
+
+_log = idaeslog.getLogger(__name__)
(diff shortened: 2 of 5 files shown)


src/watertap_contrib/seto/costing/units/med_tvc_surrogate.py
```diff
@@ -0,0 +1,258 @@
...
+    blk.heat_exchanger_specific_area = pyo.Var(
+        initialize=100,
+        units=pyo.units.m**2 / (pyo.units.kg / pyo.units.s),
+        doc="Specific heat exchanger area",
+    )
+
+    blk.thermal_storage_capacity = pyo.Var(
+        initialize=5,
+        units=pyo.units.kWh,
+        doc="Thermal storage capacity",
```
adam-a-a, line 164: This seems more like a flowsheet level variable than a unit-level one

src/watertap_contrib/seto/costing/units/med_tvc_surrogate.py
```diff
@@ -0,0 +1,258 @@
...
+    blk.med_specific_cost = pyo.Var(
+        initialize=100,
+        units=pyo.units.USD_2018 / (pyo.units.m**3 / pyo.units.day),
+        doc="MED system cost per m3/day distillate",
+    )
+
+    blk.heat_exchanger_specific_area = pyo.Var(
+        initialize=100,
+        units=pyo.units.m**2 / (pyo.units.kg / pyo.units.s),
+        doc="Specific heat exchanger area",
```
adam-a-a, line 158: Why not have this created on the unit model?

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
```diff
@@ -0,0 +1,1394 @@
...
+
+        units_meta = self.config.property_package_water.get_metadata().get_derived_units
+
+        """
+        Add system configurations
+        """
+        self.number_effects = Param(
+            initialize=12,
+            mutable=True,
+            within=Set(initialize=[8, 10, 12, 14, 16]),
```
adam-a-a, line 137: Are these the only values that are considered validated by PSA? That is, should the user only choose from these values explicitly? Can you add a comment above this explaining the Set choice and how it applied to limitations of the model?

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
```diff
@@ -0,0 +1,1394 @@
...
+            units=pyunits.K,
+            doc="Temperature increase in last effect",
+        )
+
+        self.delta_T_cooling_reject = Param(
+            initialize=-3,
+            mutable=True,
+            units=pyunits.K,
+            doc="Temperature decrease in cooling reject water",
+        )
```
adam-a-a, line 154: Why not make these variables? We could optionally fix them within this build method and then the user would have the option to unfix as desired and solve for them.

adam-a-a: Unless these are set in the surrogate model and aren't meant to be changed. However, if that were the case, maybe these parameters shouldn't be mutable.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
```diff
@@ -0,0 +1,1394 @@
...
+        ConfigValue(
+            default=False,
+            domain=In([False]),
+            description="Holdup construction flag - must be False",
+            doc="""Indicates whether holdup terms should be constructed or not.
+    **default** - False.""",
+        ),
+    )
+    CONFIG.declare(
+        "property_package_water",
```
adam-a-a, line 86:
```suggestion
        "property_package_liquid",
```

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
```diff
@@ -0,0 +1,1394 @@
...
+            description="Property package to use for feed, distillate, brine and cooling water properties",
+            doc="""Property parameter object used to define water property calculations,
+    **default** - useDefault.
+    **Valid values:** {
+    **useDefault** - use default package from parent model or flowsheet,
+    **PhysicalParameterObject** - a PhysicalParameterBlock object.}""",
+        ),
+    )
+    CONFIG.declare(
+        "property_package_steam",
```
adam-a-a, line 99:
```suggestion
        "property_package_vapor",
```

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
```diff
@@ -0,0 +1,1394 @@
...
+            bounds=(0, None),
+            units=pyunits.kWh / pyunits.m**3,
+            doc="Specific thermal power consumption (kWh/m3)",
+        )
+
+        self.gain_output_ratio = Var(
+            initialize=10,
+            bounds=(0, None),
+            units=pyunits.kg / pyunits.kg,
+            doc="Gained output ratio (kg of distillate water per kg of heating steam",
```
adam-a-a, line 345:
```suggestion
            doc="Gained output ratio (kg of distillate water per kg of heating steam)",
```

adam-a-a: In this case, it's nice that you explain the units in more detail, but in other doc strings, no need to add units in parentheses since we already assign units to the variables, unless you are adding some form of clarity as you do here.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
```diff
@@ -0,0 +1,1394 @@
...
+        @self.Constraint(doc="Thermal power requirement calculation")
+        def eq_thermal_power_requirement(b):
+            return b.thermal_power_requirement == pyunits.convert(
+                b.specific_energy_consumption_thermal
+                * b.distillate_props[0].flow_vol_phase["Liq"],
+                to_units=pyunits.kW,
+            )
+
+        # Mass flow rate
+        @self.Constraint(doc="Feed and cooling water mass flow rate (kg/s)")
```
adam-a-a, line 634: This looks more like an enthalpy balance to me. Maybe you should rename the constraint, change comment above, and change doc string if that is the case.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
```diff
@@ -0,0 +1,1394 @@
...
+            units=pyunits.kg / pyunits.s,
+            doc="Feed and cooling water mass flow rate (kg/s)",
+        )
+
+        self.feed_cool_vol_flow = Var(
+            initialize=1000,
+            bounds=(0, None),
+            units=pyunits.m**3 / pyunits.h,
+            doc="Feed and cooling water volume flow rate (m3/h)",
+        )
```
adam-a-a, line 363: Unsure why you need these. They should be on the property model as `flow_vol_phase`

adam-a-a: Ah, perhaps because you didn't feel the need to build a state block for inlet cooling water. Is that right? If so, would be good to add that to the comment above.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
```diff
@@ -0,0 +1,1394 @@
...
+            solver : str indicating which solver to use during
+                     initialization (default = None)
+
+        Returns: None
+        """
+        init_log = idaeslog.getInitLogger(blk.name, outlvl, tag="unit")
+        solve_log = idaeslog.getSolveLogger(blk.name, outlvl, tag="unit")
+
+        opt = get_solver(solver, optarg)
+        # ---------------------------------------------------------------------
```
adam-a-a, line 712: You should add a check for degrees of freedom = 0 in the initialization routine.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
```diff
@@ -0,0 +1,1394 @@
...
+        sf = iscale.get_scaling_factor(self.feed_cool_vol_flow)
+        iscale.constraint_scaling_transform(self.eq_feed_cool_vol_flow, sf)
+
+        sf = (
+            iscale.get_scaling_factor(self.cooling_out_props[0].flow_vol_phase["Liq"])
+            / 3600
+        )
+        iscale.constraint_scaling_transform(self.eq_cool_vol_flow, sf)
+
+    def _get_gain_output_ratio_coeffs(self):
```
adam-a-a, line 920: Technically, there should also be `_get_stream_table_contents()` and `_get_performance_contents()`. Later, after successfully building, scaling, initializing and solving the model, you can report results with something like `m.fs.unit.report()`.

adam-a-a: You can check out unit models on WaterTAP to compare with.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_med_tvc_surrogate.py
```diff
@@ -0,0 +1,342 @@
...
+                ("pressure", None): 101325,
+            },
+            hold_state=True,
+        )
+
+        """
+        Specify heating steam state properties
+        """
+        # Flow rate of liquid heating steam is zero
+        steam.flow_mass_phase_comp["Liq", "H2O"].fix(0)
```
adam-a-a, line 103: Hmm.. I wonder if this is something that should be handled within the unit/property models and not when the user goes to build an example.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_med_tvc_surrogate.py
```diff
@@ -0,0 +1,342 @@
...
+            hold_state=True,
+        )
+        # Release vapor mass flow rate
+        steam.flow_mass_phase_comp["Vap", "H2O"].unfix()
+
+        """
+        Specify motive steam state properties
+        """
+        # Flow rate of liquid motive steam is zero
+        motive.flow_mass_phase_comp["Liq", "H2O"].fix(0)
```
adam-a-a, line 122: same comment as above

src/watertap_contrib/seto/unit_models/surrogate/tests/test_med_tvc_surrogate.py
```diff
@@ -0,0 +1,342 @@
...
+            hold_state=True,
+        )
+        # Release vapor mass flow rate
+        motive.flow_mass_phase_comp["Vap", "H2O"].unfix()
+
+        """
+        Specify distillate flow state properties
+        """
+        # salinity in distillate is zero
+        dist.flow_mass_phase_comp["Liq", "TDS"].fix(0)
```
adam-a-a, line 139: same as above

src/watertap_contrib/seto/unit_models/surrogate/tests/test_med_tvc_surrogate.py
```diff
@@ -0,0 +1,342 @@
...
+        assert value(
+            pyunits.convert(
+                med_tvc.feed_props[0].flow_vol_phase["Liq"]
+                - med_tvc.distillate_props[0].flow_vol_phase["Liq"]
+                - med_tvc.brine_props[0].flow_vol_phase["Liq"],
+                to_units=pyunits.m**3 / pyunits.hr,
+            )
+        ) == pytest.approx(
+            feed_flow_m3_hr - dist_flow_m3_hr - brine_flow_m3_hr, rel=1e-3
+        )
```
adam-a-a, line 261: This works but you could also just check that mass in - mass out is approximately 0 (e.g., `==pytest.approx(1e-6, rel=1e-3)`)

adam-a-a (changes requested):
Looks good, but I have a few comments/suggestions for you to review.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_med_tvc_surrogate.py
kurbansitterley: I don't disagree, but- should we be doing the same for LT-MED then? @Zhuoran29 had constraints for this in the unit model (``dist.flow_mass_phase_comp["Liq", "TDS"] == 0`` and similar for other properties) that I removed in favor of this approach. 

We could also fix these in the unit model `` build`` if we don't expect them to ever change, but I am not sure we want to be doing that sort of thing.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
kurbansitterley: I these are combined feed and cooling flow? A similar approach is used in LT-MED.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
kurbansitterley: @adam-a-a Should also make these changes in LT-MED for consistency.

Zhuoran29:
Modifications have been made based on the comments.

@kurbansitterley The test failed due to the missing package of watertap.core.util.infeasible, which is called in the PV_RO model. Any idea on what's happening there?

Zhuoran29:
The consistency of LT-MED will be updated in a different PR.

adam-a-a:
> Modifications have been made based on the comments.
> 
> @kurbansitterley The test failed due to the missing package of watertap.core.util.infeasible, which is called in the PV_RO model. Any idea on what's happening there?

@Zhuoran29 my initial guess is that the `infeasible` utilities were moved in the latest WaterTAP version on main. Try changing to `watertap.core.util.model_diagnostics.infeasible`

kurbansitterley:
> @Zhuoran29 my initial guess is that the infeasible utilities were moved in the latest WaterTAP version on main. Try changing to watertap.core.util.model_diagnostics.infeasible

@adam-a-a @Zhuoran29 yes that is correct. Ben recently moved them so the import needs to change. One of the pitfalls of pointing at watertap/main.

src/watertap_contrib/seto/costing/units/med_tvc_surrogate.py
adam-a-a: @Zhuoran29 I saw you noted that the changes were made, but can you address each comment? For example, this change wasn't made, and I suppose you have some thoughts on it.

src/watertap_contrib/seto/costing/units/med_tvc_surrogate.py
```diff
@@ -0,0 +1,258 @@
...
+        initialize=5,
+        units=pyo.units.kWh,
+        doc="Thermal storage capacity",
+    )
+
+    blk.heat_exchanger_specific_area_constraint = pyo.Constraint(
+        expr=blk.heat_exchanger_specific_area
+        == pyo.units.convert(
+            med_tvc.specific_area / feed.dens_mass_phase["Liq"],
+            to_units=(pyo.units.m**2 * pyo.units.s) / pyo.units.kg,
```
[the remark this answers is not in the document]
Zhuoran29: Please see the comments above.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
Zhuoran29: I think it should be a separate PR for LT-MED consistent.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
Zhuoran29: Added at line 834 before solving the model.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
Zhuoran29: These 2 methods added.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
Zhuoran29: Yes, it is just used to calculate the flow rate of the cooling water, and a comment is added above.

Zhuoran29: > > @Zhuoran29 my initial guess is that the infeasible utilities were moved in the latest WaterTAP version on main. Try changing to watertap.core.util.model_diagnostics.infeasible
> 
> @adam-a-a @Zhuoran29 yes that is correct. Ben recently moved them so the import needs to change. One of the pitfalls of pointing at watertap/main.

Thanks, problem solved.

Zhuoran29:
> 

Thanks, problem solved.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
Zhuoran29: Test

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
Zhuoran29: Yes, it is just used to calculate the flow rate of the cooling water, and a comment is added above.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
Zhuoran29: My response was:
The range is wider but these numbers are the ones that were validated. For value like 9,11, and etc., we can do linear interpolation. Do we want to implement this feature?

Actually I'm thinking maybe the number of effect should be a config block itself. Does it make sense?

Zhuoran29 (commented):
Changes committed.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
Zhuoran29: Thanks, it makes sense.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
Zhuoran29: Doc string, constraint name and comment updated.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
Zhuoran29: Added at line 834 before solving the model.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_med_tvc_surrogate.py
Zhuoran29: Please see the comment above. The modification of LT-MED will be made in a separate PR after this one.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_med_tvc_surrogate.py
Zhuoran29: Please refer to the comment above..

src/watertap_contrib/seto/unit_models/surrogate/tests/test_med_tvc_surrogate.py
Zhuoran29: This constraint is moved to the unit model, along with other similar constraints below.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
Zhuoran29: This renaming is applied to all scripts.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
Zhuoran29: This renaming is applied to all scripts, and corresponding changes in LT-MED will be made in a separate PR.

src/watertap_contrib/seto/costing/units/med_tvc_surrogate.py
Zhuoran29: This variable is removed here and calculated in the unit model.

src/watertap_contrib/seto/costing/units/med_tvc_surrogate.py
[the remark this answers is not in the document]
Zhuoran29: Please see the comment above.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_med_tvc_surrogate.py
Zhuoran29: An additional test is added for it.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
adam-a-a: Yes, that makes sense, and I think you could go ahead and make it a config option if you like. Also, I think we can at least include the intermediate values for interpolation.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
adam-a-a: I wonder if we want to keep this flexible and make them variables that subsequently get fixed. At a minimum, it would be useful to add comments above each Param clarifying that these were parameters attributed to a particular configuration of which the surrogate model is based on.

src/watertap_contrib/seto/analysis/net_metering/PV_RO.py
```diff
@@ -38,6 +38,8 @@
 from watertap.examples.flowsheets.RO_with_energy_recovery.RO_with_energy_recovery import (
     calculate_operating_pressure,
 )
+from watertap.core.util.model_diagnostics.infeasible import *
+
```
adam-a-a, line 42:
```suggestion
```

src/watertap_contrib/seto/unit_models/surrogate/tests/test_med_tvc_surrogate.py
```diff
@@ -0,0 +1,340 @@
...
+        assert value(
+            med_tvc.brine_props[0].flow_mass_phase_comp["Liq", "TDS"]
+            - med_tvc.feed_props[0].flow_mass_phase_comp["Liq", "TDS"]
+        ) == pytest.approx(feed_mass_flow_tds - brine_mass_flow_tds, rel=1e-6)
+
+        assert value(
+            med_tvc.brine_props[0].flow_mass_phase_comp["Liq", "TDS"]
+            - med_tvc.feed_props[0].flow_mass_phase_comp["Liq", "TDS"]
+        ) == pytest.approx(0, rel=1e-3)
```
adam-a-a, line 267:
```suggestion
        ) == pytest.approx(0, abs=1e-3)
```
When checking for a 0 value, should use absolute tolerance instead of relative tolerance.

adam-a-a (approved):
LGTM- nice job! I just have one more small suggested change that I'd like you to see and implement, but approving anyway.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_med_tvc_surrogate.py
Zhuoran29: Ah sure, it makes sense.

src/watertap_contrib/seto/costing/units/med_tvc_surrogate.py
```diff
@@ -0,0 +1,218 @@
+import pyomo.environ as pyo
```
kurbansitterley, line 1: At some point we will want a header for these files. Won't let it hold up the PR.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
kurbansitterley: @Zhuoran29 do you want to implement as CONFIG option? I am ok with it either way. This is the way we did it for LT-MED though. But if you do add it as a CONFIG option, make sure to add a test for it.

But also, it the model works well (i.e. doesn't give insane numbers) for number effects not in the Set, that might be useful to include.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
Zhuoran29: Actually I think you both make sense. Physically it should be a variable, but then we want to fix them right after the declaration, and add comments about it. I think I want to do it in a separate PR.

PR #33, opened 2023-03-15
Add BasicWaterProperties prop pack

kurbansitterley:
Add very basic property package with mass concentration and volumetric flow as state variables to be used with simple pretreatment models.

### DIFF
--- a/src/watertap_contrib/seto/property_models/basic_water_properties.py
+++ b/src/watertap_contrib/seto/property_models/basic_water_properties.py
@@ -238,7 +238,7 @@ class _BasicWaterStateBlock(StateBlock):
             else:
                 self.release_state(flags)
 
-    def release_state(blk, flags, outlvl=idaeslog.NOTSET):
+    def release_state(self, flags, outlvl=idaeslog.NOTSET):
         """
         Method to release state variables fixed during initialization.
 
@@ -249,13 +249,13 @@ class _BasicWaterStateBlock(StateBlock):
                     hold_state=True.
             outlvl : sets output level of of logging
         """
-        init_log = idaeslog.getInitLogger(blk.name, outlvl, tag="properties")
+        init_log = idaeslog.getInitLogger(self.name, outlvl, tag="properties")
 
         if flags is None:
             return
 
         # Unfix state variables
-        revert_state_vars(blk, flags)
+        revert_state_vars(self, flags)
         init_log.info("State Released.")
 
 
@@ -338,14 +338,14 @@ class BasicWaterStateBlockData(StateBlockData):
             doc="Dynamic viscosity of solution",
         )
 
-    def get_material_flow_terms(blk, p, j):
-        return blk.flow_mass_comp[j]
+    def get_material_flow_terms(self, j):
+        return self.flow_mass_comp[j]
 
-    def get_enthalpy_flow_terms(blk, p):
+    def get_enthalpy_flow_terms(self, p):
         raise NotImplementedError
 
-    def get_material_density_terms(blk, p, j):
-        return blk.conc_mass_comp[j]
+    def get_material_density_terms(self, j):
+        return self.conc_mass_comp[j]
 
     def get_energy_density_terms(self, p):
         raise NotImplementedError
--- a/src/watertap_contrib/seto/property_models/basic_water_properties.py
+++ b/src/watertap_contrib/seto/property_models/basic_water_properties.py
@@ -89,8 +89,8 @@ class BasicWaterParameterBlockData(PhysicalParameterBlock):
             units=pyunits.kg / pyunits.m**3,
             mutable=True,
             doc="Mass density of flow",
-            )
-        
+        )
+
         self.visc_d = Param(
             initialize=0.001,
             units=pyunits.kg / pyunits.m / pyunits.s,
--- a/src/watertap_contrib/seto/property_models/basic_water_properties.py
+++ b/src/watertap_contrib/seto/property_models/basic_water_properties.py
@@ -383,7 +383,7 @@ class BasicWaterStateBlockData(StateBlockData):
                 try:
                     sf_c = self.params.default_scaling_factor[("conc_mass_comp", j)]
                 except KeyError:
-                    iscale.get_scaling_factor(
+                    iscale.set_scaling_factor(
                         self.conc_mass_comp[j], default=1, warning=True
                     )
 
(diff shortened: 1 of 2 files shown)


src/watertap_contrib/seto/property_models/basic_water_properties.py
```diff
@@ -0,0 +1,397 @@
...
+        )
+
+    def get_material_flow_terms(blk, p, j):
+        return blk.flow_mass_comp[j]
+
+    def get_enthalpy_flow_terms(blk, p):
+        raise NotImplementedError
+
+    def get_material_density_terms(blk, p, j):
+        return blk.conc_mass_comp[j]
```
adam-a-a, line 348: I suspect an error here if you add a test for this and try to get material density for water. Maybe you should add a conditional here to return density if j is H2O.

src/watertap_contrib/seto/property_models/basic_water_properties.py
```diff
@@ -0,0 +1,397 @@
...
+        # Check definition of solute list
+        solute_list = self.config.solute_list
+
+        for j in solute_list:
+            self.add_component(str(j), Solute())
+
+        # Define default value for mass density of solution
+        self.dens_mass_default = 1000 * pyunits.kg / pyunits.m**3
+        # Define default value for dynamic viscosity of solution
+        self.visc_d_default = 0.001 * pyunits.kg / pyunits.m / pyunits.s
```
adam-a-a, line 90: I think we could create the Params for these here (without "_default" at the end of the name). Then, for the on-demand methods, we could just use `add_object_reference()`

adam-a-a (changes requested):
@kurbansitterley I want to fast-track this one for @MuktaHardikar 's PR #40 to get working. There is just one change that is required--check my comment on `get_material_density_terms`.

As for my other comments, we can make an issue as a reminder to come back and modify as needed.

adam-a-a (approved):
Strange- it seems you had already addressed my comments, but when I looked at the code, the changes didn't show up, hence the request for changes. Looks like you took care of everything!

src/watertap_contrib/seto/property_models/basic_water_properties.py
```diff
@@ -0,0 +1,397 @@
...
+        self.set_default_scaling("visc_d", 1e3)
+
+    @classmethod
+    def define_metadata(cls, obj):
+        obj.add_default_units(
+            {
+                "time": pyunits.s,
+                "length": pyunits.m,
+                "mass": pyunits.kg,
+                "amount": pyunits.mol,
```
Zhuoran29, line 114: Just curious do we need all the basic units or we can remove "mol" if it's not used.

src/watertap_contrib/seto/property_models/basic_water_properties.py
```diff
@@ -0,0 +1,397 @@
...
+
+        for j, v in self.conc_mass_comp.items():
+            sf_c = iscale.get_scaling_factor(self.conc_mass_comp[j])
+            if sf_c is None:
+                try:
+                    sf_c = self.params.default_scaling_factor[("conc_mass_comp", j)]
+                except KeyError:
+                    iscale.get_scaling_factor(
+                        self.conc_mass_comp[j], default=1, warning=True
+                    )
```
Zhuoran29, line 388: Is here missing a line to assign the scaling factor to `self.conc_mass_comp[j]` ?
Also I think it's better to also use the multiplicative inverse of the values as the default scaling factors, instead of 1, because as tested in the test file, the concentration could vary a lot among different solutes.

src/watertap_contrib/seto/property_models/basic_water_properties.py
```diff
@@ -0,0 +1,397 @@
...
+        if j == "H2O":
+            return self.dens_mass
+        else:
+            return self.conc_mass_comp[j]
+
+    def get_energy_density_terms(self, p):
+        raise NotImplementedError
+
+    def default_material_balance_type(self):
+        return MaterialBalanceType.componentTotal
```
Zhuoran29, line 354: I wonder if this could cause error if someone tries to add material balance, because we assume the same density for the solution and solvent. (could be wrong)

Zhuoran29 (commented):
Look good to me, except I think there's a missing line.

src/watertap_contrib/seto/property_models/basic_water_properties.py
kurbansitterley: That's a good point. @adam-a-a should it be ``MaterialBalance.none`` (if that exists) since this is meant to be basic?

src/watertap_contrib/seto/property_models/basic_water_properties.py
kurbansitterley: Yes that is correct! Good catch, should be ``set_scaling_factor`` rather than ``get_scaling_factor``

MuktaHardikar (dismissed):

src/watertap_contrib/seto/property_models/basic_water_properties.py
```diff
@@ -0,0 +1,397 @@
...
+            initialize=298.15,
+            bounds=(273.15, 373.15),
+            units=pyunits.K,
+            doc="Temperature",
+        )
+
+    def _pressure(self):
+        self.pressure = Var(
+            initialize=101325,
+            bounds=(1e5, None),
```
Matthew-Boyd, line 327: Is this lower bound too restrictive (especially at altitude)?

src/watertap_contrib/seto/property_models/basic_water_properties.py
```diff
@@ -0,0 +1,397 @@
...
+                return b.flow_mass_comp[j] == b.flow_vol * b.conc_mass_comp[j]
+
+        self.eq_flow_mass_comp = Constraint(
+            self.params.component_list, rule=rule_flow_mass_comp
+        )
+
+    def _temperature(self):
+        self.temperature = Var(
+            initialize=298.15,
+            bounds=(273.15, 373.15),
```
Matthew-Boyd, line 319: Not considering solutes affecting freezing point or pressured water?

Matthew-Boyd (approved):

src/watertap_contrib/seto/property_models/basic_water_properties.py
adam-a-a: MaterialBalance.none does exist but I don't know that this is an issue just yet.

Zhuoran29 (approved):
LGTM!

PR #30, opened 2023-02-23
Add trough surrogate as unit model

kurbansitterley:
Converts work done for #20 to unit model format usable in flowsheet, associated data, and test files.

``solar_models > surrogate > trough`` contains:
- ``trough_surrogate.py`` is the unit model itself
- ``trough_surrogate.json`` is the surrogate model data
- ``test_trough_surrogate.py`` is the test file (wip)

``solar_models > surrogate > trough > data`` contains:
- ``pysam_run_trough.py`` used to generate data for surrogate
- ``training_trough_surrogate.py`` used to train surrogate
- ``trough_data.pkl`` data for surrogate
- other .jsons and weather file for PySAM

PR #34, opened 2023-03-15
Remove unused import that now has a different path in watertap

adam-a-a:
I figured I'd make this a quick PR to resolve test failures although I already addressed this issue on draft PR #30. This PR simply removes the import from `watertap.core.util.infeasible` which is now located in `watertap.core.util.model_diagnostics.infeasible`


adam-a-a:
Merging without waiting for reqs since this not a controversial change

PR #35, opened 2023-03-15
Setting up readthedocs

adam-a-a:
Addresses issue #23 

Adding @lbianchi-lbl's checklist here to get readthedocs going:
- [x]  .readthedocs.yaml file
- [x]  docs directory
- [x]  docs/conf.py
- [x]  docs/index.rst    **(sort of)**
- [x]  docs/Makefile (optional)
- [x]  Adding the Sphinx dependencies to requirements-dev.txt     **(sort of)**

PR #37, opened 2023-03-17
Add flat plate surrogate as unit model

Matthew-Boyd:
Creates a flat plate collector unit model in a format usable in flowsheets, associated data and test files

solar_models > surrogate > flat_plate contains:

    flat_plate_surrogate.py is the unit model itself
    flat_plate_surrogate.json is the surrogate model data
    test_flat_plate_surrogate.py is the test file

solar_models > surrogate > flat_plate > data contains:

    pysam_run_flat_plate.py used to generate data for surrogate
    training_flat_plate_surrogate.py used to train surrogate
    flat_plate_data.pkl data for surrogate
    other .jsons and weather file for PySAM


adam-a-a:
@Matthew-Boyd the tests are failing due to a recently merged PR in WaterTAP. My PR should fix this (PR #35 - even though its focus is on bringing in readthedocs)

adam-a-a:
OK, tests passing now after merging my PR #35 and merging main into this PR.

Matthew-Boyd:
Thanks @adam-a-a !

src/watertap_contrib/seto/costing/solar/flat_plate_surrogate.py
```diff
@@ -0,0 +1,185 @@
...
+    )
+
+    blk.variable_operating_by_generation = pyo.Var(
+        initialize=0,
+        units=costing.base_currency / (pyo.units.MWh * costing.base_period),
+        bounds=(0, None),
+        doc="Variable operating cost of flat plate plant per MWh generated",
+    )
+
+    blk.fix_all_vars()
```
adam-a-a, line 77: Actually, this line is not necessary, as it is accounted for in the decorator below.

src/watertap_contrib/seto/solar_models/surrogate/flat_plate/data/pysam_run_flat_plate.py
```diff
@@ -0,0 +1,333 @@
...
+    """
+    # 3D PLOT
+    # fig = plt.figure(figsize=(8,6))
+    # ax = fig.add_subplot(1, 1, 1, projection='3d')
+    # surf = ax.plot_trisurf(df.iloc[:,0], df.iloc[:,1], df.iloc[:,2], cmap=plt.cm.viridis, linewidth=0.2)
+    # modld_pts = ax.scatter(df.iloc[:,0], df.iloc[:,1], df.iloc[:,2], c='black', s=15)
+    # ax.set_xlabel(df.columns[0])
+    # ax.set_ylabel(df.columns[1])
+    # ax.set_zlabel(df.columns[2])
+    # plt.show()
```
adam-a-a, line 216: remove commented

src/watertap_contrib/seto/costing/solar/flat_plate_surrogate.py
Matthew-Boyd: Thanks. Removed.

src/watertap_contrib/seto/solar_models/surrogate/flat_plate/data/pysam_run_flat_plate.py
Matthew-Boyd: Removed.

src/watertap_contrib/seto/costing/solar/flat_plate_surrogate.py
```diff
@@ -0,0 +1,183 @@
...
+    )
+
+    blk.indirect_frac_direct_cost = pyo.Var(
+        initialize=0,
+        units=pyo.units.dimensionless,
+        bounds=(0, 1),
+        doc="Fraction of direct costs, including contingency, for indirect costs",
+    )
+
+    blk.tax_frac_direct_cost = pyo.Var(
```
kurbansitterley, line 56: Do you know if this fraction is consistent across different solar models we are using? It is in others and I am wondering if it makes more sense just to have it part of the system costing pkg.

kurbansitterley (changes requested):
Nothing major but a few changes should be made.

src/watertap_contrib/seto/solar_models/surrogate/flat_plate/data/pyssc.py
```diff
@@ -0,0 +1,524 @@
+import sys, os
+from ctypes import *
```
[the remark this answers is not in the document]
Matthew-Boyd: Done

src/watertap_contrib/seto/costing/solar/flat_plate_surrogate.py
Matthew-Boyd: `tax_frac_direct_cost` and `contingency_frac_direct_cost` are currently used in SAM's photovoltaic and flat plate models but not in the trough (heat) model. However, they probably should be, and thus could be here. The default values for contingency are different, but that probably doesn't need to be the case.

Please advise on where to refactor these variables to.

src/watertap_contrib/seto/solar_models/surrogate/flat_plate/flat_plate_surrogate.py
```diff
@@ -0,0 +1,275 @@
...
+        # Revert back to standard output
+        sys.stdout = oldstdout
+
+        self.dataset_filename = os.path.join(
+            os.path.dirname(__file__), "data/flat_plate_data.pkl"
+        )
+        self.n_samples = 100
+        self.training_fraction = 0.8
+
+    def _create_rbf_surrogate(
```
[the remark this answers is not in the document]
Matthew-Boyd: I like the idea, I'll implement it.

src/watertap_contrib/seto/solar_models/surrogate/flat_plate/test_flat_plate_surrogate.py
```diff
@@ -0,0 +1,187 @@
...
+    def data(self):
+        # Read data manually for repeatability instead of using the training/validation function with random splits
+        df = pd.read_pickle(DATASET_FILENAME)
+        df = df.sample(n=90, random_state=1)  # random_state ensures reproducibility
+        return {"training": df[:80], "validation": df[80:90]}
+
+    @pytest.fixture(scope="class")
+    def flat_plate_frame(self):
+        # create model and flowsheet
+        m = ConcreteModel()
```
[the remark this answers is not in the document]
Matthew-Boyd: I reformatted these tests to be similar to the trough tests.

src/watertap_contrib/seto/core/solar_energy_base.py
```diff
@@ -112,3 +121,158 @@ def calculate_scaling_factors(self):
 
         if callable(self._scaling):
             self._scaling(self)
+
+
+class SolarEnergyBase(SolarEnergyBaseData):
```
kurbansitterley, line 126: @Matthew-Boyd is there a reason you made a separate class here? Were you having import issues?

The reason I ask is I think that this is creating another class with the same name in this file after the ``declare_process_block_data`` decorator above. I like having all the surrogate methods consolidated here but they should just be in the main class.

adam-a-a: I wonder if there's value in keeping them separate so that one could be used as a base for physical or surrogate models while this new one is a base just for surrogates.

adam-a-a: @Matthew-Boyd do you have a test for this new base class?

kurbansitterley: But @adam-a-a does the ``delcare_process_block_data`` decorator not create a class called ``SolarEnergyBase`` using the ``SolarEnergyBaseData`` class? And thus having a class with the same name in the same file would create import issues? This is how all the models in WaterTAP work- the class is called like ``UnitModelData`` but then the decorator creates a class called ``UnitModel`` which is then what you import and what goes in the ``__init__.py`` files (see the LT-MED model as an example). 

Adding the methods in the above class should have the same effect.

kurbansitterley:
Shifted to active team member to continue development.

PR #38, opened 2023-03-17
Update watertap dependency to match tag pr967 

adam-a-a:
Instead of depending on watertap main, which keeps breaking our tests, i made a tag to PR #967 on the WaterTAP repo to serve as a checkpoint to depend on in the interim. That way, new changes on watertap won't break our tests, and we can adapt as needed in a more controlled fashion.


Matthew-Boyd (approved):

Zhuoran29 (approved):
Sounds a good idea to me.

PR #40, opened 2023-03-21
Chemical softening 0D model

MuktaHardikar:
Todo:
Replace constants in costing and unit model


src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
```diff
@@ -0,0 +1,833 @@
...
+    make_fixed_operating_cost_var,
+)
+
+"""
+TODO: Remove the hard coded constants in the costing equations
+      Don't fix variables in the unit model code
+"""
+
+
+class SofteningProcedureType(StrEnum):
```
kurbansitterley, line 32: You should import this class from the unit model rather than redefining it here.

kurbansitterley: And you can also just test the strings (e.g.`` if softening_type == "single_stage_lime"`` rather than ``if softening_type == SofteningProcedureType.single_stage_lime``)... but I am actually not seeing where this is used anywhere in the costing file so maybe this can be removed?

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
```diff
@@ -0,0 +1,833 @@
...
+        domain=NonNegativeReals,
+        units=blk.costing_package.base_currency,
+    )
+
+    # endregion
+
+    # region Capital cost component constraints
+
+    # Mixing tank
+    if value(blk.unit_model.vel_gradient_mix) <= 300:
```
kurbansitterley, line 236: I'm not wild about using ``value`` here because it typically evaluates at the initialized value (I think)... but interested to hear @adam-a-a thoughts.

But also- the difference in power demand between each of the velocity gradients can't be that high (correct me if I'm wrong). So maybe we just use the middle one?

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
```diff
@@ -0,0 +1,833 @@
...
+        blk.mix_tank_capital_cost_constraint = Constraint(
+            expr=blk.mix_tank_capital_cost
+            == (
+                0.0002
```
kurbansitterley, line 240: We don't want hardcoded values in WaterTAP models typically. These should be defined in ``build_chem_softening_cost_param_block`` and can be called something like "tank_cap_cost_A_coeff". And should probably be ``Params`` rather than ``Vars``.

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
```diff
@@ -0,0 +1,833 @@
...
+                    pyunits.convert(
+                        blk.unit_model.volume_mixer, to_units=pyunits.ft**3
+                    )
+                )
+                + 29756
+            )
+        )
+
+    else:
+        print("Error - Out of bounds (300 - 1000 s-1)")
```
kurbansitterley, line 312: It shouldn't be possible to get to this point. It could be a configuration check or just properly bounding the velocity gradient variable. But this goes to my earlier comment on we don't want to be using value in build functions. 

If it needs to be an error, we want to ``raise ValueError`` or something like that so that code execution stops, rather than just a print statement. Because if code execution continues ``mix_tank_capital_cost`` ends up being unconstrained.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_chemical_softening.py
```diff
@@ -0,0 +1,211 @@
...
+    unscaled_constraints_generator,
+    badly_scaled_var_generator,
+)
+import idaes.logger as idaeslog
+
+# Get default solver for testing
+solver = get_solver()
+
+
+class TestChemSoft:
```
kurbansitterley, line 48: Since you have a few different CONFIG arguments for the model (softening type, softening train type), ideally we would test all of them but we definitely want to at least test a few of them.

src/watertap_contrib/seto/unit_models/zero_order/chemical_softening_0D.py
```diff
@@ -0,0 +1,949 @@
...
+from idaes.core.util.config import is_physical_parameter_block
+from idaes.core.util.misc import StrEnum
+
+from idaes.core.util.exceptions import ConfigurationError, InitializationError
+import idaes.core.util.scaling as iscale
+import idaes.logger as idaeslog
+
+from watertap.core import InitializationMixin
+
+__author__ = "Kurban Sitterley, Abdiel Lugo"
```
kurbansitterley, line 38: Remove my name and add yours!

src/watertap_contrib/seto/unit_models/zero_order/chemical_softening_0D.py
```diff
@@ -0,0 +1,949 @@
...
+
+_log = idaeslog.getLogger(__name__)
+
+
+class SofteningTrainType(StrEnum):
```
kurbansitterley, line 43: I don't see this CONFIG argument being used anywhere in the unit model. If it isn't necessary, you should remove it.

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
```diff
@@ -0,0 +1,826 @@
+# from watertap.core.zero_order_costing import ZeroOrderCostingData
+# from idaes.core import declare_process_block_class
```
adam-a-a, line 2: Remove commented code

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
```diff
@@ -0,0 +1,826 @@
...
+        doc="Cost of MgCl2 $/kg",
+    )
+
+    blk.Na2CO3_cost = Var(
+        initialize=0.65,
+        units=costing.base_currency / pyunits.kg,
+        doc="Cost of soda ash $/kg",
+    )
+
+    blk.fix_all_vars()
```
adam-a-a, line 63: This line shouldn't be necessary.

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
```diff
@@ -0,0 +1,826 @@
...
+
+
+@register_costing_parameter_block(
+    build_rule=build_chem_softening_cost_param_block,
+    parameter_block_name="chemical_softening",
+)
+def cost_chem_softening(blk):
+
+    """
+    Capital and operating costs for excess lime and soda softening
```
adam-a-a, line 73: If this is meant to account for lime dosing as well, should there be a cost for CaO or CaOH2?

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
```diff
@@ -0,0 +1,826 @@
...
+    """
+    chem_soft = blk.costing_package.chemical_softening
+    make_capital_cost_var(blk)
+    make_fixed_operating_cost_var(blk)
+
+    blk.mixer_energy = Var(
```
adam-a-a, line 79: I've been fairly loose on people about this in WaterTAP, but since we are focusing on integration with power systems in this repo, anything with units of Watts should be named `power`, and anything with Joules or kWh should include the name `energy`. So in this case, can we rename the var as `mixer_power`?

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
```diff
@@ -0,0 +1,826 @@
...
+        units=pyunits.W,
+        domain=NonNegativeReals,
+        doc="Power consumption for rapid mix",
+    )
+
+    blk.floc_energy = Var(
+        initialize=100,
+        units=pyunits.W,
+        domain=NonNegativeReals,
+        doc="Power consumption for rapid mix",
```
adam-a-a, line 90: It looks the doc string has to be updated for all of these variables.

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
```diff
@@ -0,0 +1,826 @@
...
+    )
+
+    blk.recarb_basin_source_capital_cost = Var(
+        initialize=1e6,
+        doc="Capital cost of recarbonation tank - liquid CO2 as CO2 source",
+        domain=NonNegativeReals,
+        units=blk.costing_package.base_currency,
+    )
+
+    blk.lime_feed_system_capital_cost = Var(
```
adam-a-a, line 150: Did costs for this originate from McGiveny and Kawamura? In any case, I think it'd be nice to include the reference maybe in the doc string towards the beginning of the file (or above this var if there are various refs)

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
```diff
@@ -0,0 +1,826 @@
...
+        units=blk.costing_package.base_currency,
+    )
+
+    blk.admin_capital_cost = Var(
+        initialize=1e6,
+        doc="Capital cost of administrative, laboratory, and maintenance building",
+        domain=NonNegativeReals,
+        units=blk.costing_package.base_currency,
+    )
+    # endregion
```
adam-a-a, line 163: delete

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
```diff
@@ -0,0 +1,826 @@
...
+
+    # region Operating cost components
```
adam-a-a, line 165:
```suggestion
    # Operating cost components
```

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
```diff
@@ -0,0 +1,826 @@
...
+    )
+
+    blk.admin_op_cost = Var(
+        initialize=1e5,
+        doc="Operational cost of administrative, laboratory and maintenance building",
+        domain=NonNegativeReals,
+        units=blk.costing_package.base_currency,
+    )
+
+    # endregion
```
adam-a-a, line 231:
```suggestion
```

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
```diff
@@ -0,0 +1,826 @@
...
+
+    # region Capital cost component constraints
```
adam-a-a, line 233:
```suggestion
    # Capital cost component constraints
```

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
```diff
@@ -0,0 +1,826 @@
...
+                        blk.unit_model.volume_mixer, to_units=pyunits.ft**3
+                    )
+                )
+                ** 2
+                + 18928
+            )
+        )
+
+    else:
+        print("Error - Out of bounds (300 - 1000 s-1)")
```
adam-a-a, line 526: We would either use a logger error or warning, raise an exception, or both, but we want to avoid printing things like this (especially if meant to denote an error).

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
adam-a-a: I agree.

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
adam-a-a: Oh yeah- since `vel_gradient_mix` is a variable, you can imagine the case where the user unfixes this variable. Now, the value will be changing with every solver iteration. A potential way around this, and I guess this is slightly more advanced and has some known issues in particular circumstances, would be to use an expression if (`Expr_IF`) rather than these if-else conditionals. We can talk more about it, but another approach could be to simplify if possible for a first iteration of this model.

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
adam-a-a: It'd be interesting to see a plot of what the cost is vs `vel_gradient_mix`. Perhaps a smooth surrogate could also be created and used to replace these conditionals.

src/watertap_contrib/seto/unit_models/zero_order/chemical_softening_0D.py
```diff
@@ -0,0 +1,949 @@
...
+
+# Import IDAES cores
+from idaes.core import (
+    declare_process_block_class,
+    UnitModelBlockData,
+    useDefault,
+)
+from idaes.core.solvers.get_solver import get_solver
+from idaes.core.util.tables import create_stream_table_dataframe
+from idaes.core.util.constants import Constants
```
adam-a-a, line 28: unused import- can delete

src/watertap_contrib/seto/unit_models/zero_order/chemical_softening_0D.py
```diff
@@ -0,0 +1,949 @@
...
+                    to_units=pyunits.kg / pyunits.d,
+                )
+
+            def eq_Na2CO3_dosing(b):
+                return b.Na2CO3_dosing == pyunits.convert(
+                    (b.Ca_hardness_nonCaCO3 + b.Mg_hardness_nonCaCO3)
+                    * 1.06
+                    * prop_in.flow_vol,
+                    to_units=pyunits.kg / pyunits.d,
+                )
```
adam-a-a, line 567: delete- you add this as a constraint somewhere above

src/watertap_contrib/seto/unit_models/zero_order/chemical_softening_0D.py
```diff
@@ -0,0 +1,949 @@
...
+            {
+                "Feed Inlet": self.inlet,
+                "Liquid Outlet": self.outlet,
+                "Waste Outlet": self.waste,
+            },
+            time_point=time_point,
+        )
+
+    def _get_performance_contents(self, time_point=0):
+
```
adam-a-a, line 948: This is where you choose which variables and/or expression values would be worthwhile to show to the user. The user would run a  line such as `m.fs.chem_softening_unit.report()` and would get back values listed here. Not absolutely necessary at this stage but can be added in a subsequent PR.

src/watertap_contrib/seto/unit_models/zero_order/chemical_softening_0D.py
```diff
@@ -0,0 +1,949 @@
...
+                units=pyunits.kg / pyunits.m**3,
+                doc="Magnesium in influent converted to kg/m3 as CaCO3",
+            )
+
+        if "Alkalinity_2-" in comps:
+            self.check_alk = Expression(
+                expr=self.Ca_CaCO3 - prop_in.conc_mass_comp["Alkalinity_2-"]
+            )
+
+            if self.check_alk is NonNegativeReals:
```
adam-a-a, line 396: Should we allow for this leeway?

src/watertap_contrib/seto/unit_models/zero_order/chemical_softening_0D.py
```diff
@@ -0,0 +1,949 @@
...
+
+
+class SofteningProcedureType(StrEnum):
+    single_stage_lime = "single_stage_lime"
+    excess_lime = "excess_lime"
+    single_stage_lime_soda = "single_stage_lime_soda"
+    excess_lime_soda = "excess_lime_soda"
+
+
+@declare_process_block_class("ChemicalSoftening0D")
```
adam-a-a, line 55: I am not 100% clear. Is this a 0D model or a zero-order (assumed performance across the board) model? I ask because this file is named 0D but is in the zero_order folder. Meanwhile you also have a chem softening 0d unit in the units folder. I haven't reviewed that file yet; if they are different and this is meant to be a simpler ZO representation, please change name to `ChemicalSofteningZO`

src/watertap_contrib/seto/unit_models/surrogate/tests/test_chemical_softening.py
```diff
@@ -0,0 +1,211 @@
+import pytest
```
adam-a-a, line 1: This test file should be relocated in a test folder located in the same directory as the unit model, not in the tests folder in the surrogates directory.

src/watertap_contrib/seto/unit_models/zero_order/chemical_softening_0D.py
adam-a-a: I was about to agree but then came to the realization that this is still a draft PR, so perhaps @MuktaHardikar has some plans for this StrEnum class

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
kurbansitterley: That's true. Having them as Vars or mutable Params would also eliminate the need for all these conditionals since the user could then just fix them to whatever is appropriate for the velocity gradient they are using.

I would also like to see a plot of the different powers generated from each regression for velocity gradient. I am not convinced using one vs another would result in significant differences to LCOW or energy consumption, so leaving the option might add unnecessary complexity to a otherwise pretty simple model.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_chemical_softening.py
kurbansitterley: Also we need a header.

src/watertap_contrib/seto/unit_models/zero_order/chemical_softening_0D.py
kurbansitterley: I agree. I started this model (and am responsible for the name) with aspirations to make it more 0D-ish. But it is definitely a ZO model.

src/watertap_contrib/seto/unit_models/zero_order/chemical_softening_0D.py
MuktaHardikar: I had this here incase we wanted to add the other softening train type. But I am removing it for now.

kurbansitterley (changes requested):
I saw a few changes needed. But is generally good! How did you handle the different energy consumption for the mixers?

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
MuktaHardikar: Removed value and if statement. The coefficients of the equation in each scenario is now a mutable parameter.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_chemical_softening.py
MuktaHardikar: Tested each softening train type. I have two softening trains in the test file.

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
MuktaHardikar: Cost of chemical additions added to the costing model. References for the values also included.

src/watertap_contrib/seto/costing/units/chemical_softening_0D.py
MuktaHardikar: Reference for cost function was included in the costing block.

src/watertap_contrib/seto/unit_models/zero_order/chemical_softening_0D.py
MuktaHardikar: Added calculated chemical dosing variables to this function.

src/watertap_contrib/seto/unit_models/zero_order/chemical_softening_0D.py
MuktaHardikar: This constraint is repeated for 2 softening configs that involve addition of Na2CO3.

kurbansitterley (approved):
LGTM, thanks for making those changes

src/watertap_contrib/seto/costing/units/chemical_softening_zo.py
```diff
@@ -0,0 +1,968 @@
...
+
+    blk.purity = Param(
+        mutable=True,
+        initialize=1,
+        doc="MgCl2 purity",
+        units=pyunits.dimensionless,
+    )
+
+    costing = blk.parent_block()
+    costing.add_defined_flow("mgcl2", blk.cost / blk.purity)
```
adam-a-a, line 91: In the WT costing package, we have the following: https://github.com/watertap-org/watertap/blob/main/watertap/costing/watertap_costing_package.py#L179-L185

At first I was going to suggest conforming to what we have in WaterTap, since this is meant to be part of WaterTAP. However, we may want to incorporate what you did here to WT costing package. What do you think?

src/watertap_contrib/seto/costing/units/chemical_softening_zo.py
```diff
@@ -0,0 +1,968 @@
...
+        mutable=True,
+        doc="Coefficient in lime feed system operational cost",
+    )
+
+    blk.admin_op_exp = Param(
+        initialize=0.4589,
+        mutable=True,
+        doc="Exponent in lime feed system operational cost",
+    )
+
```
adam-a-a, line 450: Can you imagine a case where you might (in the future) want to fit some of these parameter values based on some provided set of data? If yes, then those params should be vars (and subsequently fixed). If not, keep as is. For parameter estimation, parameters have to be declared as variables to make use of IDAES' parameter estimation tool.

src/watertap_contrib/seto/costing/units/chemical_softening_zo.py
```diff
@@ -0,0 +1,968 @@
...
+        blk.fixed_operating_cost_constraint = Constraint(
+            expr=blk.fixed_operating_cost
+            == blk.mix_tank_op_cost
+            + blk.floc_tank_op_cost
+            + blk.sed_basin_op_cost
+            + blk.recarb_basin_op_cost
+            + blk.lime_feed_op_cost
+            + blk.lime_sludge_mngt_op_cost
+            + blk.admin_op_cost
+        )
```
adam-a-a, line 923: I don't see the difference between these constraints. Is the conditional needed?

src/watertap_contrib/seto/costing/units/chemical_softening_zo.py
```diff
@@ -0,0 +1,968 @@
...
+
+    mgcl2_dosing = pyunits.convert(
+        blk.unit_model.MgCl2_dosing * blk.unit_model.properties_in[0].flow_vol,
+        to_units=pyunits.kg / pyunits.year,
+    )
+
+    co2_dosing = pyunits.convert(
+        blk.unit_model.CO2_first_basin + blk.unit_model.CO2_second_basin,
+        to_units=pyunits.kg / pyunits.year,
+    )
```
adam-a-a, line 961: Any reason why these shouldn't be declared as Expressions on the model for later access, as opposed to local aliases?

src/watertap_contrib/seto/costing/units/chemical_softening_zo.py
kurbansitterley: I had never seen that before and I think it is an exception to how other chemical flows/costs are handled, for example from the IX costing method: https://github.com/watertap-org/watertap/blob/main/watertap/costing/units/ion_exchange.py

I think @bknueven changed how we did this a while go so we didn't have a bunch of chemical costs on the main costing block when most of the time they are only used for specific units. So in short - this is already how it is done in WaterTAP and I don't know how ``magnesium_chloride_cost`` slipped into the WaterTAP costing pkg in this form.

src/watertap_contrib/seto/costing/units/chemical_softening_zo.py
MuktaHardikar: I removed the conditional. Originally, MgCl2 was included as an operational cost which was in the else function. But MgCl2 is not a cost flow.

src/watertap_contrib/seto/costing/units/chemical_softening_zo.py
MuktaHardikar: Updated these to be expressions.

src/watertap_contrib/seto/costing/units/chemical_softening_zo.py
kurbansitterley: My 2cents - I don't think these would be worth using parmest for or that you'd have a situation where you couldn't fit it using Excel or something else and then just change the mutable parameter values. The regression approach would also have to match what the reference used here, and I imagine that might not be the case. Obviously it is easy to change them to vars and they are fixed via the ``@register_costing_parameter_block`` decorator, so the effort is minimal. But I am agnostic on the utility I guess.

src/watertap_contrib/seto/costing/units/chemical_softening_zo.py
MuktaHardikar: I agree. Not included chemical costs in the main costing block makes it cleaner. It is easier to track which flows are included in respective unit models and not have to go back to the main costing block and update it.

src/watertap_contrib/seto/costing/units/chemical_softening_zo.py
kurbansitterley: It is fine either way I think. The values are reportable via the ``aggregate_flow_costs`` or whatever the aggregated variable is. Thought it can be nice to directly access the costed flows via the unit model's costing block. (Also, I imagine this gives a WARNING since they are Expressions costing flows and aren't Constraints, but I don't think that change is worth making, especially for this PR).

adam-a-a:
LGTM

adam-a-a (approved):

adam-a-a:
@MuktaHardikar approved, but just went to double check if you `assert_units_consistent` and may have missed it, but didn't that check in your test for the unit. I would suggest adding it in and squeezing in another PR later if that is the case.

PR #41, opened 2023-03-23
Modify LT-MED and add interpolation for LT-MED and MED-TVC

Zhuoran29:
### Add interpolation for intermediate number of effects that doesn't have a surrogate equation
- Change parameter `number_effects` to a CONFIG block
- Expand the range of effect number in LT-MED from [3,6,9,12,14] to 3-14
- Expand the range of effect number in MED-TVC from [8,10,12,14,16] to 8-16
- Test of the interpolated results were done locally, and I don't feel the need to add a separate test for an interpolated value (so far).

### DIFF
--- a/src/watertap_contrib/seto/costing/units/lt_med_surrogate.py
+++ b/src/watertap_contrib/seto/costing/units/lt_med_surrogate.py
@@ -37,13 +37,6 @@ def build_lt_med_surrogate_cost_param_block(blk):
         doc="Fraction of capital cost for insurance",
     )
 
-    blk.cost_storage_per_kwh = pyo.Var(
-        initialize=26,
-        units=costing.base_currency / pyo.units.kWh,
-        bounds=(0, None),
-        doc="Cost of thermal storage per kWh",
-    )
-
     blk.cost_chemicals_per_vol_dist = pyo.Var(
         initialize=0.04,
         units=costing.base_currency / pyo.units.m**3,
@@ -152,31 +145,6 @@ def cost_lt_med_surrogate(blk):
         doc="MED system cost per m3/day distillate",
     )
 
-    blk.heat_exchanger_specific_area = pyo.Var(
-        initialize=100,
-        units=pyo.units.m**2 / (pyo.units.kg / pyo.units.s),
-        doc="Specific heat exchanger area",
-    )
-
-    blk.thermal_storage_capacity = pyo.Var(
-        initialize=5,
-        units=pyo.units.kWh,
-        doc="Thermal storage capacity",
-    )
-
-    blk.heat_exchanger_specific_area_constraint = pyo.Constraint(
-        expr=blk.heat_exchanger_specific_area
-        == pyo.units.convert(
-            lt_med.specific_area / feed.dens_mass_phase["Liq"],
-            to_units=(pyo.units.m**2 * pyo.units.s) / pyo.units.kg,
-        )
-    )
-
-    blk.thermal_storage_capacity_constraint = pyo.Constraint(
-        expr=blk.thermal_storage_capacity
-        == lt_med.thermal_power_requirement * lt_med_params.hours_thermal_storage
-    )
-
     blk.capacity = pyo.units.convert(
         dist.flow_vol_phase["Liq"], to_units=pyo.units.m**3 / pyo.units.day
     )
@@ -206,7 +174,7 @@ def cost_lt_med_surrogate(blk):
                 lt_med_params.cost_fraction_evaporator
                 * (
                     (
-                        blk.heat_exchanger_specific_area
+                        lt_med.specific_area_per_kg_s
                         / lt_med_params.heat_exchanger_ref_area
                     )
                     ** lt_med_params.heat_exchanger_exp
@@ -223,7 +191,6 @@ def cost_lt_med_surrogate(blk):
     blk.capital_cost_constraint = pyo.Constraint(
         expr=blk.capital_cost
         == blk.total_system_cost
-        + lt_med_params.cost_storage_per_kwh * blk.thermal_storage_capacity
     )
 
     blk.fixed_operating_cost_constraint = pyo.Constraint(
--- a/src/watertap_contrib/seto/costing/units/lt_med_surrogate.py
+++ b/src/watertap_contrib/seto/costing/units/lt_med_surrogate.py
@@ -189,8 +189,7 @@ def cost_lt_med_surrogate(blk):
     )
 
     blk.capital_cost_constraint = pyo.Constraint(
-        expr=blk.capital_cost
-        == blk.total_system_cost
+        expr=blk.capital_cost == blk.total_system_cost
     )
 
     blk.fixed_operating_cost_constraint = pyo.Constraint(
--- a/src/watertap_contrib/seto/costing/units/lt_med_surrogate.py
+++ b/src/watertap_contrib/seto/costing/units/lt_med_surrogate.py
@@ -121,12 +121,6 @@ def cost_lt_med_surrogate(blk):
     brine = lt_med.brine_props[0]
     base_currency = blk.config.flowsheet_costing_block.base_currency
 
-    blk.total_system_cost = pyo.Var(
-        initialize=100,
-        units=base_currency,
-        doc="MED system cost",
-    )
-
     blk.membrane_system_cost = pyo.Var(
         initialize=100,
         units=base_currency,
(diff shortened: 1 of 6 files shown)


src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
```diff
@@ -187,11 +192,15 @@ def eq_distillate_temp(b):
                 == b.feed_props[0].temperature + b.delta_T_last_effect
             )
 
+        # Salinity in distillate is zero
+        @self.Constraint(doc="distillate salinity")
+        def eq_distillate_salinity(b):
+            return b.distillate_props[0].flow_mass_phase_comp["Liq", "TDS"] == 0
```
adam-a-a, line 198: You could also fix to zero and skip constraint, but if tests are passing, it might be OK for now.

src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
```diff
@@ -356,354 +376,388 @@ def temp_last_effect(b):
...
 
+        # Surrogate equations were built for 3,6,9,12,14 effects
+        # For intermediate number of effects (4,5,7,8,10,11,13), linear interpolation is adopted
+        @self.Constraint(doc="Gain output ratio surrogate equation")
+        def eq_gain_output_ratio(b):
+            if b.config.number_effects in [3, 6, 9, 12, 14]:
+                return b.gain_output_ratio == _get_gain_output_ratio(
+                    b.config.number_effects
+                )
+            else:  # number_effects in [4,5,7,8,10,11,13]
```
adam-a-a, line 710: Should have another conditional that prevents <3 effects and >14 effects. We can raise an exception to be more stringent about the surrogate model's valid range.

src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
Zhuoran29: Yes it makes sense. This constraint is removed.

src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
Zhuoran29: Sure, the exceptions are added.

src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
```diff
@@ -356,354 +374,396 @@ def temp_last_effect(b):
...
+                    interp_point2 - b.config.number_effects
+                ) / (interp_point2 - interp_point1) + _get_gain_output_ratio(
+                    interp_point2
+                ) * (
+                    b.config.number_effects - interp_point1
+                ) / (
+                    interp_point2 - interp_point1
+                )
+            else:
+                raise Exception(
```
kurbansitterley, line 726: Should raise ``ConfigurationError`` and be checked at the beginning of ``build()`` I'd say

src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
```diff
@@ -1050,7 +1154,7 @@ def _get_gain_output_ratio_coeffs(self):
             ],
             9: [
                 1.67e-06,
-                5.9507846,
+                5.9507846 + 1e-3,
```
kurbansitterley, line 1157: What is the reason for this opposed to just having ``5.9517846``?

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
```diff
@@ -404,240 +401,287 @@ def eq_feed_to_cooling_isobaric(b):
...
             )
 
         heating_steam_mass_flow_rate_coeffs = (
             self._get_heating_steam_mass_flow_rate_coeffs()
         )
 
-        @self.Constraint(doc="heating steam mass flow rate surrogate equation")
-        def eq_heating_steam_mass_flow_rate(b):
+        # Heating steam mass flow rate surrogate equation, as a function of the number of effect
+        def _get_heating_steam_mass_flow_rate(num_effect):
```
kurbansitterley, line 491: Typically these "hidden" functions (those prepended with underscore  _) are at the end of the file. And it's my preference also... saves a lot of over-scrolling when I am looking through Constraints :)

src/watertap_contrib/seto/unit_models/surrogate/tests/test_lt_med_surrogate.py
```diff
@@ -43,35 +43,131 @@ def LT_MED_frame(self):
         m.fs.water_prop = SeawaterParameterBlock()
```
kurbansitterley, line 43: Should have a short build test for the ``ConfigurationError`` in both these test files.

kurbansitterley (changes requested):
Mostly is good but I made some minor comments that should be addressed before merging.

src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
```diff
@@ -108,6 +109,15 @@ class LTMEDData(UnitModelBlockData):
     see property package for documentation.}""",
         ),
     )
+    CONFIG.declare(
+        "number_effects",
+        ConfigValue(
+            default=12,
+            domain=In([i for i in range(3, 15)]),
+            description="Number of effects of the MED_TVC system",
+            doc="""A ConfigBlock specifying the number of effects, which should be an integer between 8 to 16.""",
```
adam-a-a, line 118:
```suggestion
            doc="""A ConfigBlock specifying the number of effects, which should be an integer between 8 to 16.""",
```
```suggestion
            doc="""A ConfigBlock specifying the number of effects, which should be an integer between 3 to 14.""",
```

src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
adam-a-a: Since you already of the domain set for number_effects, you should add a test to check what happens (whether you get exception as expected and what the error message is). You could then modify as needed to raise a relevant exception

src/watertap_contrib/seto/costing/units/lt_med_surrogate.py
```diff
@@ -221,9 +189,7 @@ def cost_lt_med_surrogate(blk):
     )
 
     blk.capital_cost_constraint = pyo.Constraint(
-        expr=blk.capital_cost
-        == blk.total_system_cost
-        + lt_med_params.cost_storage_per_kwh * blk.thermal_storage_capacity
+        expr=blk.capital_cost == blk.total_system_cost
```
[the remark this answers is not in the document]
Zhuoran29: Thanks for spotting this. This equation included the battery cost before I removed it. Since it's gone, I removed `total_system_cost` as well.

src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
Zhuoran29: I added a test for the `ValueError` that is raised by IDAES `ConfigBlock` as Adam pointed out that it should be handled by IDAES already, but I also raised a `ConfigurationError` in the `build()` function as well (it shouldn't be necessary in the current setup as it won't be triggered).

src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
Zhuoran29: Thanks for noticing it! It's a change I did for an initialization problem and forgot to remove it.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
Zhuoran29: It makes sense! I moved them to the end of the script.

src/watertap_contrib/seto/unit_models/surrogate/med_tvc_surrogate.py
```diff
@@ -404,240 +401,287 @@ def eq_feed_to_cooling_isobaric(b):
...
+                return (
+                    b.gain_output_ratio
+                    == (
+                        _get_gain_output_ratio(b.config.number_effects - 1)
+                        + _get_gain_output_ratio(b.config.number_effects + 1)
+                    )
+                    / 2
+                )
+            else:
+                raise Exception(
```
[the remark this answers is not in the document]
Zhuoran29: Please see comments above.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_lt_med_surrogate.py
Zhuoran29: A test for the `ValueError` raised by IDAES was added.

adam-a-a (changes requested):
Minor change suggested (change domain so that more useful `ConfigurationError `can be raised during build)

adam-a-a (approved):
LGTM

PR #44, opened 2023-04-06
Steady-state VAGMD_batch unit model

Zhuoran29:
### Steady-state VAGMD_batch unit model

- This unit model uses surrogate equations to simulate the membrane performance during each batch cycle, and aggregate the material and energy flow for the overall performance of the system
- This model needs an external equation-oriented function to calculate the number of batch cycles required. A surrogate expression of this function might be needed in the future for the optimization of an integrated system.

#### TODO:
- Add tests for different CONFIG options
- Add initialization and report

### DIFF
--- a/src/watertap_contrib/seto/unit_models/surrogate/tests/test_vagmd_batch_surrogate.py
+++ b/src/watertap_contrib/seto/unit_models/surrogate/tests/test_vagmd_batch_surrogate.py
@@ -149,7 +149,7 @@ class TestVAGMD:
             high_brine_salinity = False
         )
 
-        m.fs.vagmd.RR.fix(RR)
+        # m.fs.vagmd.recovery_vol_phase["Liq"].fix(RR)
         m.fs.vagmd.feed_props[0].conc_mass_phase_comp["Liq","TDS"].fix(S0)
         m.fs.vagmd.feed_props[0].temperature.fix(Ttank+273.15)
         m.fs.vagmd.feed_props[0].flow_vol_phase["Liq"].fix(FFR/3600/1000)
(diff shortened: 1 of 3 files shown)


src/watertap_contrib/seto/unit_models/surrogate/tests/test_vagmd_batch_surrogate.py
```diff
@@ -0,0 +1,484 @@
...
+        assert not m.fs.vagmd.config.dynamic
+        assert not m.fs.vagmd.config.has_holdup
+        assert m.fs.vagmd.config.property_package is m.fs.properties
+
+    @pytest.mark.unit
+    def test_build(self, VAGMD_frame):
+        m = VAGMD_frame
+
+        # test ports
+        port_lst = ["feed", "permeate"]
```
adam-a-a, line 131: third port for brine? I am guessing the surrogate can skip permeate inlet

src/watertap_contrib/seto/unit_models/surrogate/tests/test_vagmd_batch_surrogate.py
```diff
@@ -0,0 +1,484 @@
...
+            feed_flow_rate,
+            temp_cond_in,
+            feed_salinity,
+            temp_tank,
+            module_type,
+            final_salinity,
+        )
+        permeate_flow_rate_init = (
+            permeate_flux_init * Area
+        )  # Initial permeate flow rate (L/h)
```
adam-a-a, line 78: If you put the comment above this line rather than at the end of it, black won't reformat it this way.

Issue #47, opened 2023-04-13
Change domain so that more useful `configurationerror` can be raised during build


adam-a-a:
change domain so that more useful `configurationerror` can be raised during build

_Originally posted by @adam-a-a in https://github.com/watertap-org/watertap-seto/pull/41#discussion_r1162927732_

adam-a-a:
In reference to LTMED and MEDTVC models

Zhuoran29:
Issue resolved in PR#49.

Issue #45, opened 2023-04-11
Refine LTMED initialization for 10/11 effects


adam-a-a:
We'll want to generalize the initialization routine to handle this. That is, for interpolated number of effects = 10 or 11, initialization should be refined so that this case initializes successfully.

In the LTMED test (#41), a conditional has to be introduced to check that number of effects = 11, and a new initial value must be provided to specific area in order for the model to initialize successfully.

adam-a-a:
A temporary solution is in place in #49 but we should keep this issue up as a reminder for a more permanent solution at some point.

adam-a-a:
This issue was resolved with PR #59

PR #49, opened 2023-04-17
Correction of LT_MED and minor changes

Zhuoran29:
- Correct the surrogate equation of specific area for LT-MED at 3 effects
- Refine LT_MED initialization routine for 11 effects regarding issue #45 
- Change domain of config option `number_effects` to raise `ConfigurationError` regarding issue #47

### DIFF
--- a/src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
+++ b/src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
@@ -127,7 +127,7 @@ class LTMEDData(UnitModelBlockData):
         # Check if the number of effects is valid
         if self.config.number_effects not in [i for i in range(3, 15)]:
             raise ConfigurationError(
-                "The number of effects should be an integer between 3 to 14"
+                f"The number of effects was specified as {self.config.number_effects}. The number of effects should be specified as an integer between 3 to 14."
             )
 
         """
--- a/src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
+++ b/src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
@@ -23,7 +23,7 @@ from pyomo.environ import (
     check_optimal_termination,
     units as pyunits,
 )
-from pyomo.common.config import ConfigBlock, ConfigValue, In
+from pyomo.common.config import ConfigBlock, ConfigValue, In, PositiveInt
 
 # Import IDAES cores
 from idaes.core import (
@@ -113,7 +113,7 @@ class LTMEDData(UnitModelBlockData):
         "number_effects",
         ConfigValue(
             default=12,
-            domain=In([i for i in range(3, 15)]),
+            domain=PositiveInt,
             description="Number of effects of the LT_MED system",
             doc="""A ConfigBlock specifying the number of effects, which should be an integer between 3 to 14.""",
         ),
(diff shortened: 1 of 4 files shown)


src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
```diff
@@ -113,7 +113,7 @@ class LTMEDData(UnitModelBlockData):
         "number_effects",
         ConfigValue(
             default=12,
-            domain=In([i for i in range(3, 15)]),
+            domain=PositiveInt,
```
kurbansitterley, line 116: Would this allow for number effects >15 or <3?

Zhuoran29: Yes. We would like to raise a more explicit error message when inputting invalid values. The imbedded `ValueError `from ConfigBlock is a big Vague, by just saying "invalid value for configuration 'number_effects'".

The more straightforward message was added at line 129.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_lt_med_surrogate.py
```diff
@@ -116,8 +117,8 @@ def test_num_effects_domain(self):
         m.fs.liquid_prop = SeawaterParameterBlock()
         m.fs.vapor_prop = WaterParameterBlock()
 
-        error_msg = "invalid value for configuration 'number_effects'"
-        with pytest.raises(ValueError, match=error_msg):
+        error_msg = "The number of effects should be an integer between 3 to 14"
```
adam-a-a, line 120:
```suggestion
        error_msg = f"The number of effects was specified as {object_pointing_to_num_effects}. The number of effects should be specified as an integer between 3 to 14."
```

adam-a-a (changes requested):
Recommended a little clarification on the error message

kurbansitterley (approved):
LGTM provided Adam's changes pass.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_lt_med_surrogate.py
Zhuoran29: Changes made accordingly for both models.

src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
```diff
@@ -635,6 +635,14 @@ def initialize_build(
         # Check degree of freedom
         assert degrees_of_freedom(blk) == 0
 
+        # For 10 and 11 effects, specific area may need a new initialization value,
+        # because the surrogate eqn for 9 and 12 effects are largely different
+        if blk.config.number_effects == 11:
+            blk.specific_area_per_m3_day.value = 3
```
adam-a-a, line 641: I think this is something we should still keep an issue on as a reminder and revisit, in case there is a more elegant solution to this, rather than a hard-coded value subject to a particular number of effects. Also, is it not an issue for 10 effects?

Zhuoran29: Yea it is not elegant and I'll reconsider it. Will need to see how the model perform on the whole solution space by visualizing it as Kurby did for 12 effects.

PR #56, opened 2023-06-15
Modify LT_MED and PV_RO for Pyomo 6.6.1 compatibility

Zhuoran29:
#### Updated the scaling factors so that the following models are compatible with both Pyomo 6.5.0 and 6.6.1:
- Modified the scaling factor of 1 constraint in LT_MED
- Added the scaling factors for 3 constraints in reverse_osmosis_0D model that is used in PV_RO

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ setup(
     python_requires=">=3.8",
     install_requires=[
         "watertap @ https://github.com/watertap-org/watertap/archive/refs/tags/pr967.zip",
-        # "pyomo <= 6.5",
+        "pyomo == 6.6.1",
         "pytest >= 7",
         "nrel-pysam == 3.0.2",
     ],
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ setup(
     python_requires=">=3.8",
     install_requires=[
         "watertap @ https://github.com/watertap-org/watertap/archive/refs/tags/pr967.zip",
-        "pyomo == 6.6.1",
+        "pyomo <= 6.6.1",
         "pytest >= 7",
         "nrel-pysam == 3.0.2",
     ],
--- a/setup.py
+++ b/setup.py
@@ -9,7 +9,7 @@ setup(
     python_requires=">=3.8",
     install_requires=[
         "watertap @ https://github.com/watertap-org/watertap/archive/refs/tags/pr967.zip",
-        "pyomo <= 6.5",
+        # "pyomo <= 6.5",
         "pytest >= 7",
         "nrel-pysam == 3.0.2",
     ],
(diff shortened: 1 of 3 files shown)


setup.py
```diff
@@ -9,7 +9,7 @@
     python_requires=">=3.8",
     install_requires=[
         "watertap @ https://github.com/watertap-org/watertap/archive/refs/tags/pr967.zip",
-        "pyomo <= 6.5",
+        # "pyomo <= 6.5",
```
adam-a-a, line 12: For the final merge of this PR, we should specify the actual version of Pyomo that this is running against, i.e., 
`pyomo <= 6.6.1` (or whatever version it is we're looking at).

Zhuoran29:
@adam-a-a Yea, I cannot debug for PV_RO without having a Linux system at this moment - seems missing scaling factors for RO module. Good news is LT_MED works in both windows and linux now.

setup.py
Zhuoran29: Pyomo version is specified now.

Aslo I noticed there's other packages that have a different version from the previous environment, for example `pytest_cov == 4.1.0` vs. previous 4.0.0. Not sure if those packages will change the solvability of the model, but `pyomo` should be the major one.

adam-a-a: I think that corresponds to pytest's code coverage and wouldn't expect it to change solvability of the model. @lbianchi-lbl would know best about any caveats related to that in general though.

setup.py
```diff
@@ -9,7 +9,7 @@
     python_requires=">=3.8",
     install_requires=[
         "watertap @ https://github.com/watertap-org/watertap/archive/refs/tags/pr967.zip",
-        "pyomo <= 6.5",
+        "pyomo == 6.6.1",
```
adam-a-a, line 12:
```suggestion
        "pyomo <= 6.6.1",
```
Did this not work?

Zhuoran29: It works but I was trying to be specific with the version.

adam-a-a:
> @adam-a-a Yea, I cannot debug for PV_RO without having a Linux system at this moment - seems missing scaling factors for RO module. Good news is LT_MED works in both windows and linux now.

I doubt it has to do with the RO module specifically. A lot of those warnings you are probably seeing about missing scaling factors has to do with some changes that were made in Pyomo and how IDAES has passed suffixes (but didn't explore details of this).

Zhuoran29:
> I think RO module should be fine as well. I meant that the PV RO flowsheet didn't call the calculate scaling factors function from RO unit model during initialization. I added that in PV RO and it worked in Windows now but not in Linux. Probably there's other scaling factors that need to be added as well.
> 
> 
> 
> 
> 
> > @adam-a-a Yea, I cannot debug for PV_RO without having a Linux system at this moment - seems missing scaling factors for RO module. Good news is LT_MED works in both windows and linux now.
> 
> 
> 
> I doubt it has to do with the RO module specifically. A lot of those warnings you are probably seeing about missing scaling factors has to do with some changes that were made in Pyomo and how IDAES has passed suffixes (but didn't explore details of this).

Zhuoran29:
Closed with scaling issue resolved by PR #59

PR #58, opened 2023-07-11
Update Solar Energy Surrogates

kurbansitterley:
- Add ``FlatPlateSurrogate``, test file, costing, and data for surrogate
- Update to ``SolarEnergyBase`` class and units that inherit from it
- Add test file for ``SolarEnergyBase``  and test data for test file


zacharybinger (approved):

src/watertap_contrib/seto/core/solar_energy_base.py
```diff
@@ -112,3 +120,205 @@ def calculate_scaling_factors(self):
...
+        self.trainer = PysmoRBFTrainer(
+            input_labels=self.input_labels,
+            output_labels=self.output_labels,
+            training_dataframe=self.data_training,
+        )
+
+        # Set PySMO options
+        self.trainer.config.basis_function = "gaussian"  # default = gaussian
+        self.trainer.config.solution_method = "algebraic"  # default = algebraic
+        self.trainer.config.regularization = True  # default = True
```
adam-a-a, line 180: I wonder if these should be config options or at least more accessible for the user to change as desired.

src/watertap_contrib/seto/solar_models/surrogate/trough/test_trough_surrogate.py
```diff
@@ -174,7 +182,6 @@ def test_validation(self, trough_frame):
         for row in data["validation"].itertuples():
             m.fs.trough.heat_load.fix(row.heat_load)
             m.fs.trough.hours_storage.fix(row.hours_storage)
```
adam-a-a, line 184: Just a note for the record- after removing the usage of default ipopt and using ipopt-watertap in `test_validation`, IPOPT output shows that the search direction becomes too small and stops at 3 or 4 iterations; this happens when `hours_of_storage` is fixed to 1 but may not be limited to that (e.g., also noticed same issue when `hours_of_storage` was fixed to 2 but didn't look beyond that). However, with regular ipopt, these problematic cases solve in 1 iteration, so the problem has to do with one of the customized settings of ipopt-watertap. However, one could argue that the problem lies within the surrogate equations since solving should be seamless for both versions of ipopt. @bknueven this could be worth exploring to determine whether ipopt-watertap should undergo some adjustment (or not) considering this as a test case.

kurbansitterley:
Closed for re-factoring/updating in future PR.

Issue #51, opened 2023-04-27
Trough surrogate model infeasible solves with WaterTAP solver


kurbansitterley:
### Description 

The trough surrogate model does not consistently solve with ``solver = get_solver()`` but consistently solves with ``solver = SolverFactory("ipopt")``. Sometimes the infeasibilities are with only the trough on the flowsheet and sometimes when the trough costing package is added. 

### Steps to Reproduce

The most straightforward way to reproduce is to replace the solver in the trough test file with the WaterTAP solver. The tests  with and without costing will fail. The current trough test file does not have scaling, but if the following scaling is added to the test frame, the costing test will pass but the validation test will still fail:

```
set_scaling_factor(m.fs.trough.electricity, 1e-4)
set_scaling_factor(m.fs.trough.heat, 1e-5)
set_scaling_factor(m.fs.trough.heat_annual, 1e-9)
set_scaling_factor(m.fs.trough.electricity_annual, 1e-8)
```

Here is a specific build with LT-MED included on the flowsheet that will results in an infeasible solve:

```
from pyomo.environ import (
    ConcreteModel,
    assert_optimal_termination,
    units as pyunits,
    SolverFactory,
)

from idaes.core.solvers import get_solver
from idaes.core.util.scaling import *
from idaes.core import FlowsheetBlock, UnitModelCostingBlock
from idaes.core.util.model_statistics import *

from watertap.property_models.seawater_prop_pack import SeawaterParameterBlock
from watertap.property_models.water_prop_pack import WaterParameterBlock
from watertap_contrib.seto.costing import (
    EnergyCosting,
    SETOSystemCosting,
    TreatmentCosting,
)

from watertap_contrib.seto.solar_models.surrogate.trough import TroughSurrogate
from watertap_contrib.seto.unit_models.surrogate import LTMEDSurrogate
from watertap.core.util.model_diagnostics.infeasible import *

solver_wt = get_solver()
solver_ipopt = SolverFactory("ipopt")

trough_scale_cost_vars = ['fs.energy.trough.costing.capital_cost',
 'fs.energy.trough.costing.variable_operating_cost',
 'fs.energy.trough.costing.fixed_operating_cost',
 'fs.energy.trough.costing.direct_cost',
 'fs.energy.costing.aggregate_capital_cost',
 'fs.energy.costing.aggregate_fixed_operating_cost',
 'fs.energy.costing.aggregate_variable_operating_cost',
[...]

PR #59, opened 2023-07-15
Mods on LTMED and Trough Surrogate

adam-a-a:
- adds initialization to trough surrogate
- fixes report() error with LT-MED
- solve trough surrogate with watertap ipopt (in trough surrogate test file)
- remove all constraint scaling transforms on LT-MED and revise initialization

### DIFF
--- a/src/watertap_contrib/seto/solar_models/surrogate/trough/trough_surrogate.py
+++ b/src/watertap_contrib/seto/solar_models/surrogate/trough/trough_surrogate.py
@@ -21,7 +21,7 @@ import numpy as np
 from io import StringIO
 import matplotlib.pyplot as plt
 
-from pyomo.environ import Var, Constraint, Suffix, units as pyunits
+from pyomo.environ import Var, Constraint, Suffix, units as pyunits, check_optimal_termination
 
 from idaes.core import declare_process_block_class
 import idaes.core.util.scaling as iscale
@@ -30,6 +30,12 @@ from idaes.core.surrogate.pysmo_surrogate import PysmoRBFTrainer, PysmoSurrogate
 from idaes.core.surrogate.sampling.data_utils import split_training_validation
 
 from watertap_contrib.seto.core import SolarEnergyBaseData
+from idaes.core.solvers.get_solver import get_solver
+from idaes.core.util.exceptions import InitializationError
+
+import idaes.logger as idaeslog
+
+_log = idaeslog.getLogger(__name__)
 
 __author__ = "Matthew Boyd, Kurban Sitterley"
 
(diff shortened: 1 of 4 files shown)


src/watertap_contrib/seto/unit_models/surrogate/tests/test_lt_med_surrogate.py
```diff
@@ -370,7 +379,7 @@ def test_interp_values(self, number_effects):
             "flow_mass_phase_comp", 1e-2, index=("Liq", "H2O")
         )
         m.fs.liquid_prop.set_default_scaling(
-            "flow_mass_phase_comp", 1e3, index=("Liq", "TDS")
```
adam-a-a, line 373: I think this was the source of some of your scaling problems, @Zhuoran29 . After revising this, I didn't need constraint scaling anymore at all, and the issue with having to set specific area value for 10 effects during initialization was resolved too.

Zhuoran29: Thanks for identifying the issue! It makes sense just by looking at it. I'll test and do the same for MED-TVC, then commit to the same PR. Will update this comment after I do so.

src/watertap_contrib/seto/unit_models/surrogate/lt_med_surrogate.py
```diff
@@ -687,86 +683,6 @@ def calculate_scaling_factors(self):
...
-        iscale.constraint_scaling_transform(self.eq_cool_vol_flow, sf)
-
-        for t in self.flowsheet().config.time:
-            sf = iscale.get_scaling_factor(self.feed_props[t].pressure)
-            iscale.constraint_scaling_transform(
-                self.eq_feed_to_distillate_isobaric[t], sf
-            )
-            iscale.constraint_scaling_transform(self.eq_feed_to_brine_isobaric[t], sf)
-            iscale.constraint_scaling_transform(self.eq_feed_to_cooling_isobaric[t], sf)
-
```
Zhuoran29, line 769: Just want to confirm that, is the scaling of constraints not needed because all variables are already scaled?

adam-a-a: Pretty much-, I think the general guidance was to scale variables and leave the constraints as they will be handled by ipopt, unless there are some problematic constraints identified, in which case you could add some targeted contraint scaling transforms. In this case, I found that none were needed, at least for the cases tested.

adam-a-a: Furthermore, as I was probing @kurbansitterley 's issue (#51), my attempts seemed to improve when I took away constraint scaling from LT-MED.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_lt_med_surrogate.py
adam-a-a: Actually, I think it's better to just do on a separate PR.

Zhuoran29: Ok, sure.

Zhuoran29 (approved):
LGTM.

kurbansitterley (approved):
LGTM, thanks for looking into this more.

src/watertap_contrib/seto/solar_models/surrogate/trough/trough_surrogate.py
```diff
@@ -143,8 +155,48 @@ def calculate_scaling_factors(self):
             sf = iscale.get_scaling_factor(self.electricity, default=1e-3, warning=True)
             iscale.set_scaling_factor(self.electricity, sf)
 
-    def initialize_build(self):
-        pass
+    def initialize_build(
```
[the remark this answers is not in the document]
kurbansitterley: And I will take the same approach for the other surrogates.

PR #50, opened 2023-04-20
VAGMD-batch multiperiod model

Zhuoran29:
### VAGMD batch operation simulation using IDAES MultiPeriod Class. Files include:

`VAGMD_surrogate.py`: A unit model for VAGMD performance at a certain period
`VAGMD_batch_flowsheet.py`: A flowsheet for VAGMD at a certain period, collecting aggregative information and creating placeholders for variables from previous period
`VAGMD_batch_flowsheet_multiperiod.py`: Create a MultiPeriod Class to simulate a multiperiod case for batch operation

#### TODO: 

- [x] - Fix remaining scaling issue. 
- [x] - Complete test file for multiperiod class.
- [x] - Add test for unit model and flowsheet.

PR #61, opened 2023-07-20
Upload vagmd unit model PR 1/4

Zhuoran29:
Implementation of the base VAGMD unit model:

`VAGMD_surrogate_base.py`: A unit model for the performance of a single Aquastill MD module
`test_vagmd_surrogate_base.py`: Corresponding test file.

The main changes from the previous version is calculating the status of brine flow and adding a port to the brine state block.

This PR incorporates the comments from the closed PR #50.

This PR will be the first PR for VAGMD models:

1. Base model for a single MD module
2. Steady state unit model for MD
3. Multiperiod model on batch-MD
4. Multiperiod model on semi-batch LT-MED and MD


src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
```diff
@@ -0,0 +1,1233 @@
...
+        # because only volumetric flow rate is known and the temperature will be released and calculated afterwards
+        self.cooling_out_props.calculate_state(
+            var_args={
+                # Cooling water volumetric flow rate, fixed to 1265L/h, to maintain vacuum pressure inside the MD module around 200 mbar
+                ("flow_vol_phase", "Liq"): pyunits.convert(
+                    1265 * pyunits.L / pyunits.h, to_units=pyunits.m**3 / pyunits.s
+                ),
+                ("conc_mass_phase_comp", ("Liq", "TDS")): 50,
+                ("temperature", None): 25 + 273.15,
+                ("pressure", None): 101325,
```
adam-a-a, line 440: These should be Params. Also, what if we wanted to scale up and, say, assume multiple modules in parallel? Wouldn't flowrate change?

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
```diff
@@ -0,0 +1,1233 @@
...
+        self.add_port(name="feed", block=self.feed_props)
+        self.add_port(name="brine", block=self.evaporator_out_props)
+
+        """
+        Constraint equations
+        """
+        # Cooling system specification
+        if self.config.cooling_system_type == "closed":
+
+            @self.Constraint(doc="Calculate cooling power requirment")
```
adam-a-a, line 488:
```suggestion
            @self.Constraint(doc="Calculate cooling power requirement")
```

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
```diff
@@ -0,0 +1,1233 @@
...
+                )
+
+            @self.Constraint(
+                doc="Calculate inlet cooling water temperature in closed cooling mode"
+            )
+            def eq_cooling_in_temp(b):
+                return (
+                    b.cooling_in_props[0].temperature
+                    == b.feed_props[0].temperature
+                    - pyunits.convert(b.cooling_power_thermal, to_units=pyunits.W)
```
adam-a-a, line 506: Any reason why this is in Watts and kW used in constraint above?

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
```diff
@@ -0,0 +1,1233 @@
...
+            + coefficients[5] * salinity**2
+        )
+
+    """
+    Surrogate equations for membrane performance
+    """
+
+    def _get_membrane_performance(self, TEI, FFR, TCI, SgL):
+        # Model parameters
+        PFluxAS26 = [
```
adam-a-a, line 957: Seems like local assignment of these parameters aren't all necessary since these are for both modules. Technically each group of parameters be created based on which module was specified, right?

src/watertap_contrib/seto/unit_models/surrogate/tests/test_vagmd_surrogate.py
```diff
@@ -0,0 +1,511 @@
...
+        # Closed circuit, in which TCI is forced to be constant and the cooling water temperature can be adjusted.
+        if cooling_system_type == "closed":
+            m.fs.vagmd.condenser_in_props[0].temperature.fix(cond_inlet_temp + 273.15)
+        # Open circuit, in which cooling is available at a constant water temperature and condenser inlet temperature varies.
+        else:  # "open"
+            m.fs.vagmd.cooling_in_props[0].temperature.fix(cooling_inlet_temp + 273.15)
+
+        return m
+
+    @pytest.mark.unit
```
adam-a-a, line 111: I might've missed this because reviewing from phone, but do you have a test for assert_units_consistent()?

adam-a-a (commented):
I had a few comments to consider. Overall, looks good.

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
```diff
@@ -0,0 +1,1233 @@
...
+
+        # Add ports
+        self.add_port(name="feed", block=self.feed_props)
+        self.add_port(name="brine", block=self.evaporator_out_props)
+
+        """
+        Constraint equations
+        """
+        # Cooling system specification
+        if self.config.cooling_system_type == "closed":
```
kurbansitterley, line 486: Maybe this should be a ``Enum`` config also?

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
```diff
@@ -0,0 +1,1233 @@
...
+                    + pyunits.convert(b.cooling_power_thermal, to_units=pyunits.W)
+                    / b.thermal_resistance_cold
+                )
+
+        # Set alias for state properties
+        FFR = pyunits.convert(
+            self.feed_props[0].flow_vol_phase["Liq"], to_units=pyunits.L / pyunits.h
+        )
+        S = self.feed_props[0].conc_mass_phase_comp["Liq", "TDS"]
+        TEI = self.evaporator_in_props[0].temperature
```
kurbansitterley, line 559: I am not wild about acronyms as aliases from a readability perspective, but it is ok, just include in a comment what the acronyms stand for.

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
```diff
@@ -0,0 +1,1233 @@
...
+        def eq_thermal_resistance_cold(b):
+            return b.thermal_resistance_cold == pyunits.convert(
+                b.cooling_in_props[0].flow_vol_phase["Liq"]
+                * b.avg_cooling_props[0].dens_mass_phase["Liq"]
+                * b.avg_cooling_props[0].cp_mass_phase["Liq"],
+                to_units=pyunits.W / pyunits.K,
+            )
+
+        @self.Constraint(doc="Calculate the number of transfer units")
+        def eq_number_transfer_units(b):
```
kurbansitterley, line 695: I assume this doesn't solve to an integer. Any issues with that or how do you consider that? Or maybe its ok if it is a fraction?

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
```diff
@@ -0,0 +1,1233 @@
...
+            + coefficients[3] * flow_rate * salinity
+            + coefficients[4] * flow_rate**2
+            + coefficients[5] * salinity**2
+        )
+
+    """
+    Surrogate equations for membrane performance
+    """
+
+    def _get_membrane_performance(self, TEI, FFR, TCI, SgL):
```
kurbansitterley, line 955: Similar as above- just define what the acronyms are in a comment so a reader doesn't need to hunt around.

kurbansitterley (changes requested):
A few minor things to do here before merging. But this seems like a monster of a model so great job in general

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
```diff
@@ -0,0 +1,1233 @@
...
+        Model parameters
+        """
+        self.pump_efficiency = Param(
+            initialize=0.6, units=pyunits.dimensionless, doc="Pump efficiency"
+        )
+
+        self.heat_exchanger_area = Param(
+            initialize=1.34,
+            units=pyunits.m**2,
+            doc="Effective heat transfer coefficient",
```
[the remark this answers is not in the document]
Zhuoran29: Corrected.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_vagmd_surrogate.py
```diff
@@ -0,0 +1,511 @@
...
+
+        final_brine_salinity = feed_salinity / (1 - recovery_ratio)  # = 200 g/L
+        if module_type == "AS7C1.5L" and final_brine_salinity > 175.3:
+            cooling_system_type = "closed"
+            feed_flow_rate = 1100  # L/h
+            evap_inlet_temp = 80  # deg C
+            cond_inlet_temp = 25  # deg C
+            high_brine_salinity = True
+        else:
+            high_brine_salinity = False
```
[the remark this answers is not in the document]
Zhuoran29: Good suggestions. Changes applied.

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
```diff
@@ -0,0 +1,1233 @@
...
+        )
+
+        """
+        Model parameters
+        """
+        self.pump_efficiency = Param(
+            initialize=0.6, units=pyunits.dimensionless, doc="Pump efficiency"
+        )
+
+        self.heat_exchanger_area = Param(
```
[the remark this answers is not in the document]
Zhuoran29: This unit model is specialized for one MD module, and many design parameters are fixed (such as cooling water flow rate as well).

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
```diff
@@ -0,0 +1,1233 @@
...
+        )
+
+        """
+        Output variables
+        """
+        self.thermal_power = Var(
+            initialize=10,
+            bounds=(0, None),
+            units=pyunits.kW,
+            doc="Thermal power requirment",
```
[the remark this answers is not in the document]
Zhuoran29: Corrected.

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
Zhuoran29: Similar to my response above that this unit model is for 1 module only. I've considered to make it suitable for scaling up but I gave it up for the following reasons:
1. The surrogate equations use the properties of water within one module. If allowing scaling up, the state blocks will need to be duplicated for the actual flows, which is fine but could looks a bit chunky. 
2. My initial thought on the MD applications would be batch/semi-batch operation (the mode that was delivered for SETO project), because once-through process has very low recovery ratio. So I didn't expect the unit model to be used alone, but to be integrated in a multi-period structure, which considers the capacity and cost in the multi-period level. 

Of course someone might want to use it along for some reason, in that case, maybe we can have another unit model for that, and this model will be like "reverse_osmosis_base" that are used in 'oaro' or 'ro_0D'.

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
Zhuoran29: Corrected.

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
Zhuoran29: The parameter 'thermal resistance' in the equation uses Watts in its unit.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_vagmd_surrogate.py
Zhuoran29: It will fail for surrogate equations. Same in LT-MED and MED-TVC.

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
```diff
@@ -0,0 +1,1233 @@
...
+            raise Exception(
+                f"Degrees of freedom of {self.name} is {degrees_of_freedom(self)}"
+                f"before the initialization of VAGMD unit model."
+            )
+
+        opt = get_solver(solver, optarg)
+
+        with idaeslog.solver_log(solve_log, idaeslog.DEBUG) as slc:
+            res = opt.solve(self, tee=slc.tee)
+            print_close_to_bounds(self)
```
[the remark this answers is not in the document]
Zhuoran29: Thanks. I forgot to remove it.

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
```diff
@@ -0,0 +1,1233 @@
...
+    UnitModelBlockData,
+    useDefault,
+)
+from idaes.core.util.config import is_physical_parameter_block
+import idaes.core.util.scaling as iscale
+from idaes.core.solvers import get_solver
+import idaes.logger as idaeslog
+
+# Import Watertap packages
+from watertap.core.util.model_diagnostics.infeasible import *
```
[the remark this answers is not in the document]
Zhuoran29: Yep, deleted.

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
Zhuoran29: Yes. It's just a thermodynamic coefficient, but with a confusing name. It is okay to be a fraction.

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
```diff
@@ -0,0 +1,1233 @@
...
+            initialize=100,
+            bounds=(0, None),
+            units=pyunits.mbar,
+            doc="Feed flow pressure drop",
+        )
+
+        self.log_mean_temp_dif = Var(
+            initialize=0,
+            bounds=(0, None),
+            units=pyunits.C,
```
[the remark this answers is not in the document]
Zhuoran29: Corrected.

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
Zhuoran29: Comments added.

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
Zhuoran29: Comment added.

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
```diff
@@ -0,0 +1,1233 @@
...
+                b.evaporator_out_props[0].conc_mass_phase_comp["Liq", "TDS"]
+                == b.feed_props[0].conc_mass_phase_comp["Liq", "TDS"]
+                * b.feed_props[0].flow_vol_phase["Liq"]
+                / b.evaporator_out_props[0].flow_vol_phase["Liq"]
+            )
+
+        # make brine pressure at 1 bar
+        @self.Constraint(doc="brine pressure")
+        def eq_brine_pressure(b):
+            return b.evaporator_out_props[0].pressure == 101325 * pyunits.Pa
```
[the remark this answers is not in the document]
Zhuoran29: Applied.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_vagmd_surrogate.py
```diff
@@ -0,0 +1,511 @@
...
+        final_brine_salinity = feed_salinity / (1 - recovery_ratio)  # = 70 g/L
+        if module_type == "AS7C1.5L" and final_brine_salinity > 175.3:
+            cooling_system_type = "closed"
+            feed_flow_rate = 1100  # L/h
+            evap_inlet_temp = 80  # deg C
+            cond_inlet_temp = 25  # deg C
+            high_brine_salinity = True
+        else:
+            high_brine_salinity = False
+
```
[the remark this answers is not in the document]
Zhuoran29: Yes it makes sense. It does feel odd to determine the configs manually in the tests. 

However, this unit model is (almost) always used in a multiperiod flowsheet where the CONFIGs should be determined externally to the model. After a few trials and tuning with the multiperiod structure, I believe the best way is to keep the current logic. To make the test file look more reasonable, I created a helper function in the unit model that can handle this logic automatically.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_vagmd_surrogate.py
```diff
@@ -0,0 +1,511 @@
...
+        final_brine_salinity = feed_salinity / (1 - recovery_ratio)  # = 200 g/L
+        if module_type == "AS7C1.5L" and final_brine_salinity > 175.3:
+            cooling_system_type = "closed"
+            feed_flow_rate = 1100  # L/h
+            evap_inlet_temp = 80  # deg C
+            cond_inlet_temp = 25  # deg C
+            high_brine_salinity = True
+        else:
+            high_brine_salinity = False
```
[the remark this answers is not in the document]
Zhuoran29: Please see comments above.

src/watertap_contrib/seto/unit_models/surrogate/tests/test_vagmd_surrogate.py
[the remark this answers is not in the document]
Zhuoran29: Please see comments above.

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
```diff
@@ -0,0 +1,1233 @@
...
+        )
+
+        self.cooling_flow_pressure_drop = Param(
+            initialize=170, units=pyunits.mbar, doc="Cooling flow pressure drop"
+        )
+
+        """
+        MD Module type and corresponding area
+        """
+        if self.config.module_type == "AS7C1.5L":
```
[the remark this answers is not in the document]
Zhuoran29: Yep! Adopted.

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
Zhuoran29: Adopted!

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
```diff
@@ -0,0 +1,1233 @@
...
+            doc="Material properties of permeated water",
+            **tmp_dict,
+        )
+        # salinity in permeate is zero
+        self.permeate_props[0].flow_mass_phase_comp["Liq", "TDS"].fix(0)
+
+        # permeate temperature is assumed the same as feed water
+        @self.Constraint(doc="distillate temperature")
+        def eq_distillate_temp(b):
+            return b.permeate_props[0].temperature == b.feed_props[0].temperature
```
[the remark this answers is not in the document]
Zhuoran29: I wonder if we want do `.fix()` for input values, here the feed temperature on the right hand side of the equation is an input variable. We probably want to leave it free for optimization purpose. For now I leave it untouched.

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate.py
Zhuoran29: Not sure if I understood this comment. Is it trying to make the script more concise? If so, I could be wrong but I don't see a clear way to do so.

src/watertap_contrib/seto/unit_models/surrogate/vagmd_surrogate_base.py
```diff
@@ -0,0 +1,1264 @@
...
+    ):
+        # This function rewrites the operation parameters and cooling type
+        # when the brine salinity is high (>175.3 g/L) for module AS7C1.5L
+        if module_type == ModuleType.AS7C15L and high_brine_salinity:
+            feed_flow_rate = 1100  # L/h
+            evap_inlet_temp = 80  # deg C
+            cond_inlet_temp = 25  # deg C
+            self.config.cooling_system_type = CoolingType.closed
+            cooling_system_type = CoolingType.closed
+
```
adam-a-a, line 940: if module_type == ModuleType.AS7C15L:
            feed_flow_rate = 1100  # L/h
            evap_inlet_temp = 80  # deg C
            cond_inlet_temp = 25  # deg C
            final_brine_conc = 200 # kg/m3
            self.config.cooling_system_type = CoolingType.closed
            cooling_system_type = CoolingType.closed

Zhuoran29: I see. I get it we don't want to have boolean input from the users if it can be solved internally. The logic was that only when the brine salinity is high for module AS7, should we fix the operational variables to nominal point, so the if condition needs to include at least the brine salinity. Considering the actual brine salinity is not known for each time step in the multiperiod structure, I used the boolean variable to indicate the scenario in general.

So I suggest we revisit it when the PR for multiperiod is up.

adam-a-a: That sounds good to me.

adam-a-a (approved):
LGTM - this is 1 PR out 4. Any further refinement needed may be identified in subsequent PRs

kurbansitterley (approved):
LGTM, thanks for making changes

PR #62, opened 2023-08-03
Update trough surrogate

dguittet:
1. Make two separate surrogate models for different heat loads: [10, 100] and [100, 500]
2. Both of these surrogate models are scaled so the Trough unit model is updated the handle the scaling
3. Updated all tests and surrogate validation tests fixed to pass with WaterTap solver
    - costing test was modified since the initial Trough model in the test is now 100 MW heat_load instead of 1000


dguittet:
@kurbansitterley @MuktaHardikar Here's the PR for the updated surrogates. I'm not able to request reviewers, but I wanted to alert you to test these changes, especially within flowsheets containing the surrogate model.

src/watertap_contrib/seto/solar_models/surrogate/trough/test_trough_surrogate.py
```diff
@@ -113,105 +99,142 @@ def test_build(self, trough_frame):
...
+    def test_validation_large(self, trough_large_heat_load):
 
-        m = trough_frame
-        data = get_data()
+        m = trough_large_heat_load
+        data = get_data(m.fs.trough.heat_load.bounds)
         heat_annual_list = []
         electricity_annual_list = []
 
+        solver = SolverFactory("ipopt")
```
adam-a-a, line 140: @MuktaHardikar I recall you saying that these revisions made the model more stable and able to solve under previous conditions where the model was difficult to solve. I thought that you were referring to using "ipopt-watertap" rather than "ipopt". Could you clarify whether solver convergence issues still occur with "ipopt-watertap"?

MuktaHardikar: I've been using mainly ipopt-watertap but haven't checked with ipopt in the flowsheet. Previously, without the scaling in the surrogate, the model depended a lot on scaling in the flowsheet to solve. With the updates, the models default scaling is pretty good for most cases and I didn't have convergence issues with the ipopt-watertap solver.

src/watertap_contrib/seto/solar_models/surrogate/trough/trough_surrogate.py
```diff
@@ -53,16 +56,36 @@ class TroughSurrogateData(SolarEnergyBaseData):
...
+    CONFIG.declare("heat_load_range", ConfigValue(
+        domain=ListOf(float),
+        default=[10, 100],
+        description="Range of the Heat Load input",
+        doc="""Heat load range [min, max]. Options are: [10, 100] or [100, 500]"""
+    ))
+
+    surrogate_scaling_dict = {
+        (10, 100): 1e-8,
+        (100, 500): 1e-9
```
MuktaHardikar, line 68: I ended up using (10,200) and had to update this dict to run the training model and trough surrogate. I wonder if this could be updated to be more flexible?

src/watertap_contrib/seto/solar_models/surrogate/trough/data/training_trough_surrogate.py
```diff
@@ -154,23 +157,32 @@ def plot_training_validation(
...
-            plt.savefig("/plots/parity_residual_plots.png")
+            plt.savefig(Path(__file__).parent / "plots" / "parity_residual_plots.png")
             plt.close()
 
 
 #########################################################################################################
 if __name__ == "__main__":
-    dataset_filename = join(dirname(__file__), "trough_data.pkl")
-    surrogate_filename = join(dirname(__file__), "trough_surrogate_testing.json")
+    heat_load_range = (10, 100)         # must be (10, 100) or (100, 500)
```
MuktaHardikar, line 166: Maybe a comment here guiding the user to update the range here and the scaling dict in trough_surrogate?

src/watertap_contrib/seto/solar_models/surrogate/trough/test_trough_surrogate.py
adam-a-a: So technically, if we delete any occurences of `SolverFactory("ipopt")` in this PR, checks should still pass. I thought I had resolved this issue in a previous PR but had overlooked occurrences of `SolverFactory("ipopt")`. Hopefully, this PR will still pass when we remove.

src/watertap_contrib/seto/solar_models/surrogate/trough/trough_surrogate.py
adam-a-a: I agree. Since we are building in functionality to create surrogates, I wonder if we can have a default set for heat_load and storage (e.g., corresponding to the values hard-coded here in the tuples), but allow a user to choose their own bounds for the surrogate which would subsequently replace the values here.

adam-a-a: If we did that though, it'd probably make the workflow a lot riskier for new users who don't know what they're doing, so we might want appropriate safeguards in place as well as warnings for the user to be aware of, if we go this route.

dguittet: I agree this current proposed way of describing multiple surrogate models is clunky, but what is the intended workflow? How many surrogate models do we need for a single technology? I guess it would help to understand first what is the magnitude of any benefits to breaking up the design space into multiple models.

Regardless, I can try to put the scaling factor into the surrogate JSON directly so the data is included in the model info, rather than being separated like it is now, and maybe add some logic for the trough to choose the most appropriate surrogate model from the available ones given the input bounds.

dguittet:
@kurbansitterley The tests are failing due to 

```
FAILED src/watertap_contrib/seto/analysis/tests/test_PV_RO.py::TestPVRO::test_solution - assert 1.74 ± 1.7e-01 == 1.5062410112212197
  comparison failed
  Obtained: 1.50624101122[121](https://github.com/watertap-org/watertap-seto/actions/runs/5955472237/job/16154193112?pr=62#step:6:122)97
  Expected: 1.74 ± 1.7e-01
```

I don't think this PR impacts the PV_RO flowsheet so I'm not sure why the test would change. Any ideas?

adam-a-a:
> @kurbansitterley The tests are failing due to
> 
> ```
> FAILED src/watertap_contrib/seto/analysis/tests/test_PV_RO.py::TestPVRO::test_solution - assert 1.74 ± 1.7e-01 == 1.5062410112212197
>   comparison failed
>   Obtained: 1.50624101122[121](https://github.com/watertap-org/watertap-seto/actions/runs/5955472237/job/16154193112?pr=62#step:6:122)97
>   Expected: 1.74 ± 1.7e-01
> ```
> 
> I don't think this PR impacts the PV_RO flowsheet so I'm not sure why the test would change. Any ideas?

@dguittet I think the change that "broke" the test for PV_RO was made in the SolarEnergyBase class. It was a subtle change, but apparently change electricity bounds from (None, None) to a domain of NonNegativeReals triggered a reduction in LCOW:
https://github.com/watertap-org/watertap-seto/blob/980adc18ec67cec344cfdf05560182ceafd33505/src/watertap_contrib/seto/core/solar_energy_base.py#L77

dguittet:
Plan is to develop surrogate models just for the current use case via the external surrogate fitting script. Since the fitting process may be fairly bespoke and require iterations, automation within the unit model build process isn't appropriate. The surrogate model will be integrated with the unit model during the initialization process, and this should be flexible enough to allow users to choose whichever surrogate model they want to use. 

The surrogate model fitting and integration within the unit model will be tested with two locations.

kurbansitterley:
removed the PV+RO test so this is moot now

kurbansitterley (approved):

src/watertap_contrib/seto/solar_models/surrogate/trough/trough_surrogate.py
kurbansitterley: I am ok with this implementation for now. Really, only advanced users (i.e., us) will be using these surrogates for the time being. As we get better at building/using surrogates, we can probably implement some of this functionality. But for now, having to edit the bounds manually for a specific use case seems ok.

src/watertap_contrib/seto/solar_models/surrogate/trough/data/training_trough_surrogate.py
kurbansitterley: I'd view this as more a demonstration rather than a final implementation. Ultimately we will have to make new surrogates anyway for a different location.

kurbansitterley:
This looks great and I am pleased it works with WaterTAP solver, which will hopefully ease some of the solving challenges I was having.

src/watertap_contrib/seto/solar_models/surrogate/trough/data/training_trough_surrogate.py
```diff
@@ -73,8 +77,18 @@ def create_rbf_surrogate(
     return rbf_surr
 
 
-def get_training_validation(dataset_filename, n_samples, training_fraction):
+def get_training_validation(
+    dataset_filename, n_samples, training_fraction, heat_load_range, hours_storage_range
+):
```
adam-a-a, line 82: Should we have some check on heat_load_range and hours_storage_range? What happens if the user enters something invalid? Do we have a test to check that as well?

adam-a-a (commented):
From a quick scan, I am wondering if we need to add any safeguards (e.g., raise exception) if the user does something wrong when trying to supply his/her own heat_load and heat_storage ranges to train the surrogate. I also think we should have a test or two to test this functionality as well as test any safeguards.

src/watertap_contrib/seto/solar_models/surrogate/trough/data/training_trough_surrogate.py
kurbansitterley: This is a good question. I would defer to @dguittet , but I am not sure we need a test for this.

I don't think this file is tested directly. I view this as sort of a demo/how-to for creating a trough surrogate. Presumably, the user would know the limits of their system/surrogate, something that we could not know a priori for a test (unless there are practical limits to these variables, but that is a different consideration maybe).

To me, this really relates to how we want to implement surrogate generation moving forward, and how flexible and automated we want that implementation to be moving forward.

adam-a-a: As the code stands now, it looks like we're trying to grant a tad more flexibility in that the user can specify their own ranges specifically for heat load and heat storage hours if they so choose to, rather than always being bound to specific ranges that were defined for a single location. Of course, they'd be doing this at their own risk, and our documentation should reflect that down the line. 

However, since we have the functionality already in play here, whether the test is implemented directly or indirectly through the `test_trough_surrogate file`, my only suggestion is that we test what happens when using this function with an (1) alternative set of define ranges, simply to check that the functionality performs as intended (rather than going through the full workflow to create a new surrogate), and (2) add something in this function that will raise an exception if the user provides input that is unexpected and invalid, and test that this exception is raised when entering something nonsensical as input.

adam-a-a: I should add one thing--I was assuming that the function here would be intended for use to arrive at alternative surrogates while also serving as an example. Not clear whether this was meant to serve as both or simply to just serve as a representative example.

adam-a-a:
Notes from meeting: @dguittet will add some simple testing of `get_training_validation()`. That should be enough to approve and merge shortly!

src/watertap_contrib/seto/solar_models/surrogate/trough/data/training_trough_surrogate.py
dguittet: Added an error check

dguittet:
@adam-a-a All checks have passed. Could you please review to enable merging?

adam-a-a (approved):
LGTM - I will create an issue to capture any thoughts for refinement in future PRs.

PR #64, opened 2023-08-14
Mg crystallizer

Zhuoran29:
This PR includes a unit model of a crystallizer that precipitates Mg and Ca after NaCl has been removed from the WaterTAP crystallizer model.


adam-a-a:
Quick observation- it appears that you're not using a property model relevant to the components you are considering. Functionally, the model probably works, but technically, it deviates from IDAES/WaterTAP standards- i.e., separating properties from unit models. 
I can imagine a couple reasons why you would take that approach, but wanted to hear your thoughts on it.

Zhuoran29:
> Quick observation- it appears that you're not using a property model relevant to the components you are considering. Functionally, the model probably works, but technically, it deviates from IDAES/WaterTAP standards- i.e., separating properties from unit models. I can imagine a couple reasons why you would take that approach but want to hear your thoughts on it.

Yea, that's one thing I wanted to discuss as well. I'm using `cryst_prop_pack` for this model at this point because:

- Connecting ports requires both state blocks having the same property package, and this model should always follow a WaterTAP crystallizer model. (Of course we can skip port connection if using different property packages.
- Currently this model is only operable at 25 C and 1 bar (changing temperature almost changes everything in the system and this condition is believed the best, but we still probably want to include that freedom as well); so technically there's no need to use a property package specifically.
- Missing data on the density of the multi-salt composition: To incorporate the related material in the property package, this model requires at least the density of the solution, but the relationship is unknown at this point. Good news is the concentration of Mg and Ca is much smaller compared to NaCl, thus I'm using NaCl solution density as approximation.

It would definitely be within the interests for next steps to try to extend the conditions of the system and establish a property package accordingly.

adam-a-a:
Suppose you did use, say, the MCAS property package to include other components and wanted to use in combination with the NaCl prop pack. You could use a TranslatorBlock to connect NaCL with MCAS prop pack (just FYI). 

I get what you mean--that at this stage, there isn't much need for a new property package nor enough info to make it useful. However,my concern is that when WaterTAP-SETO is compared with WaterTAP and IDAES, formulations won't be consistent because of property and unit models not utilizing the separation that we often boast about as a way to grant flexibility. We could think about how to use MCAS or a similar approach instead of blending properties together with the unit.

Zhuoran29:
Sure, I agree. I don't like the property variables standing alone in the model as well. I'll see what I can do with MCAS prop pack, thanks for the info.

src/watertap_contrib/seto/unit_models/zero_order/mg_crystallizer_zo.py
```diff
@@ -0,0 +1,595 @@
...
+            initialize=0.76071,
+            units=pyunits.g / pyunits.L,
+            doc="Solubility of hydromagnesite [Mg5(CO3)4(OH)2 4H2O] at 25 deg C",
+        )
+
+        self.calcite_solubility = Param(
+            initialize=0.17631,
+            units=pyunits.g / pyunits.L,
+            doc="Solubility of calcite [CaCO3] at 25 deg C",
+        )
```
adam-a-a, line 162: These should be variables that get fixed within the model, leaving the option open for parameter estimation later on.

src/watertap_contrib/seto/unit_models/zero_order/mg_crystallizer_zo.py
```diff
@@ -0,0 +1,595 @@
...
+
+        self.Mg_molar_weight = Param(
+            initialize=24.305, units=pyunits.g / pyunits.mol, doc="Molar weight of Mg"
+        )
+
+        self.Ca_molar_weight = Param(
+            initialize=40.087, units=pyunits.g / pyunits.mol, doc="Molar weight of Mg"
+        )
```
adam-a-a, line 170: These are OK as Params since we would never fit these with parameter estimation. However, if using MCAS, you could provide these up front when creating an instance of MCAS.

PR #70, opened 2023-09-29
Trough surrogate test tolerance for linux

dguittet:
In the Linux CI, there are random test failures for `test_create_rbf_surrogate_large ` in `test_trough_surrogate.py`:
    1. https://github.com/watertap-org/watertap-seto/actions/runs/6342581771/job/17228658039
    2. https://github.com/watertap-org/watertap-seto/actions/runs/6341627645/job/17225691888

The test asserts that the surrogate model can predict the electricity within 5% and this works most of the time, except for the problems in Linux where the prediction error can go up to around ~16%. I'm not sure what's changing in the surrogate fit function on Linux that causes this. 

I modified the test by adding checks of the R2 and RMSE for the surrogate, and increasing the tolerance of the prediction error to 20%.

### DIFF
--- a/src/watertap_contrib/seto/solar_models/surrogate/trough/test_trough_surrogate.py
+++ b/src/watertap_contrib/seto/solar_models/surrogate/trough/test_trough_surrogate.py
@@ -159,10 +159,8 @@ class TestTrough:
         expected_electricity_annual_test = (
             data["validation"]["electricity_annual"] * surrogate_scaling
         )
-        tol = 5e-2
-        if platform == "linux":
-            tol = 2e-1
 
+        tol = 2e-1
         assert list(test_output["heat_annual_scaled"]) == pytest.approx(
             expected_heat_annual_test.tolist(), tol
         )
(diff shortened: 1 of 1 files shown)


dguittet:
Hi @lbianchi-lbl have you seen issues before around the performance of surrogate model fitting via PySMO being unstable on Linux?

lbianchi-lbl:
> Hi @lbianchi-lbl have you seen issues before around the performance of surrogate model fitting via PySMO being less consistent on Linux?

Good question... I don't really have knowledge of PySMO beyond occasional maintenance needs within IDAES/idaes-pse. I'd suggest opening an issue or discussion in IDAES/idaes-pse (there's a [PySMO label](https://github.com/IDAES/idaes-pse/labels/PySMO)). I'd also mention that (unfortunately) PySMO is not being funded currently AFAIK, which understandably limits the amount and extend of user support.

adam-a-a:
@dguittet we usually see these discrepancies between OS platforms due to scaling issues....so it could have something to do with that.

adam-a-a:
I think this could serve as a bandaid for now, and I'll add an issue to track this problem.

adam-a-a (approved):
LGTM - added an issue (#71) to track a long-term fix

dguittet:
@adam-a-a Ok that tracks, as the scaling is a bit bad for the electricity output. Let me try to adjust it

kurbansitterley (approved):
LGTM! Thanks for looking into this @dguittet

PR #76, opened 2023-09-29
Add automatic scaling for trough surrogates

### LINKED ISSUE #71: Surrogate results differ on Linux
Hi @lbianchi-lbl have you seen issues before around the performance of surrogate model fitting via PySMO being less consistent on Linux?

_Originally posted by @dguittet in https://github.com/watertap-org/watertap-seto/issues/70#issuecomment-1741141370_
            
Also from @dguittet:

> In the Linux CI, there are random test failures for test_create_rbf_surrogate_large  in test_trough_surrogate.py:
> 1. https://github.com/watertap-org/watertap-seto/actions/runs/6342581771/job/17228658039
> 2. https://github.com/watertap-org/watertap-seto/actions/runs/6341627645/job/17225691888
> 
> The test asserts that the surrogate model can predict the electricity within 5% and this works most of the time, except for the problems in Linux where the prediction error can go up to around ~16%. I'm not sure what's changing in the surrogate fit function on Linux that causes this.
> 
> I modified the test by adding checks of the R2 and RMSE for the surrogate, and increasing the tolerance of the prediction error to 20%.

dguittet:
In the Linux CI, there are random test failures for test_create_rbf_surrogate_large  in test_trough_surrogate.py:
1. https://github.com/watertap-org/watertap-seto/actions/runs/6342581771/job/17228658039
2. https://github.com/watertap-org/watertap-seto/actions/runs/6341627645/job/17225691888

Automatic scaling of the training data was added so that whatever subset of input data is used, the training data will be normalized by 1. The stored surrogate models of `trough_surrogate_100_500.json` and `trough_surrogate_10_100.json` is removed since the surrogate model is auto-generated.

The surrogate model fitting by PySMO is now more predictive and stable so the error tolerance for `test_create_rbf_surrogate_large ` is now 10%. Usually the error is quite a bit lower, like 5-7%, but since each instance has a different model fit, a margin was needed and hopefully 10% is enough.

fixes #71


dguittet:
@adam-a-a @kurbansitterley This is ready for review. The only test failure is due to readthedocs, but this PR didn't impact anything relevant to the docs.

adam-a-a (approved):
LGTM - one thing I noticed is that running the test will result in generating the trough surrogate json files that were deleted in this PR. We probably should add a line or two to cleanup those files, but that can be handled in a subsequent PR.

kurbansitterley (approved):
LGTM

PR #77, opened 2023-09-29
Rename SETO files/classes to REFLO

kurbansitterley:
Anywhere "seto"/"SETO" is in a file, change to "reflo"/"REFLO".

Only deviation is ``SETOWaterTAPCosting`` I changed to ``REFLOCosting`` to be a bit more concise.


adam-a-a (approved):
LGTM - something strange was going on with the tests. A couple of them were seemingly timing out for unknown reasons. I reran them a few times for them to work. Additionally, one of the surrogate tests that we thought was settled in #76 also raised an error, so I slightly adjust the relative tolerance for a regression test to make that pass.

Zhuoran29 (approved):

PR #85, opened 2023-11-30
AirStripping0D and AirWaterEquilibrium Property Model + Docs

kurbansitterley:
This PR will add:

- AirWaterEquilibrium (AWE) property model and related tests
- Docs for AWE
- AirStripping0D unit model and related tests
- Docs for air stripping model
- Docs for basic water property package


src/watertap_contrib/reflo/property_models/air_water_equilibrium_properties.py
```diff
@@ -0,0 +1,1805 @@
...
+            initialize=0.083899,
+            units=pyunits.dimensionless,
+            doc="Collision function zeta equation - F parameter",
+        )
+
+        self.collision_function_zeta_param_G = g = Param(
+            initialize=-0.011491,
+            units=pyunits.dimensionless,
+            doc="Collision function zeta equation - G parameter",
+        )
```
adam-a-a, line 1329: Conventionally, I think these params would typically go on the param block. Any particular reason why you did it this way, other than convenience while developing? 

I suppose the only reason it matters is that one would expect to access parameters from the param block, and that'd be pretty consistent with most things in IDAES/WaterTAP, if my recollection is correct.

adam-a-a: I guess the other advantage of doing it this way is that parameters associated with on-demand props wouldn't be built unnecessarily, and building these on the param block would be less straight-forward if you want to also build on-demand.

kurbansitterley: Both of those reasons are pretty much it. 

And I suppose I'd push back a bit on this convention.  I don't see the advantage/logic behind putting them on the param block and them _always_ being built, but maybe I am missing something. It makes more sense to me to build property-specific params when the on-demand property is built. I would liken this to how we started putting unit-specific costing parameters only in unit-specific costing packages so they weren't always built and cluttering up calls to ``.display()`` or sifting through components via ``.component_objects()`` calls (maybe I am the only one that does that regularly though...)

But I definitely won't die on this hill. I am not opposed to putting them on the param block (or making them Vars, which appears to also be convention).

kurbansitterley:
@dguittet any idea what is going on with these failing ``test_trough_surrogate`` tests?

kurbansitterley:
> @dguittet any idea what is going on with these failing `test_trough_surrogate` tests?

Nevermind, not clear if it has anything to do with the surrogates. Same test fail as issue 89

src/watertap_contrib/reflo/property_models/air_water_equilibrium_properties.py
adam-a-a: I won't die on this hill either. There are mountains to climb! =)

src/watertap_contrib/reflo/costing/units/air_stripping.py
```diff
@@ -0,0 +1,563 @@
...
+        units=base_currency,
+        doc="Air blower cost",
+    )
+
+    blk.blower_power = pyo.Var(
+        initialize=50,
+        bounds=(0, None),
+        units=pyo.units.kilowatt,
+        doc="Air blower power requirement",
+    )
```
adam-a-a, line 390: I would prefer this to be on the unit model itself. There could be the case where the user might only care for energy-related modeling without inclusion of costs.

src/watertap_contrib/reflo/costing/units/air_stripping.py
```diff
@@ -0,0 +1,563 @@
...
+        units=base_currency,
+        doc="Water pump cost",
+    )
+
+    blk.pump_power = pyo.Var(
+        initialize=50,
+        bounds=(0, None),
+        units=pyo.units.kilowatt,
+        doc="Water pump power requirement",
+    )
```
adam-a-a, line 376: I had a similar comment for blower_power; I think any Var that could hypothetically be used without cost (e.g., energy/power consumption, etc.) should live on the unit model rather than in it's costing module.

src/watertap_contrib/reflo/costing/units/air_stripping.py
```diff
@@ -0,0 +1,563 @@
...
+                * ax_params.blower_efficiency
+            )
+            * (
+                (prop_out.pressure / ax_params.pressure_ambient)
+                ** ax_params.power_blower_exponent
+                - 1
+            ),
+            to_units=pyo.units.kilowatt,
+        )
+    )
```
adam-a-a, line 560: move to unit model

src/watertap_contrib/reflo/property_models/air_water_equilibrium_properties.py
```diff
@@ -0,0 +1,1843 @@
...
+
+
+class LiqDiffusivityCalculation(Enum):
+    none = auto()
+    HaydukLaudie = auto()
+
+
+class VapDiffusivityCalculation(Enum):
+    none = auto()
+    WilkeLee = auto()
```
adam-a-a, line 102: Would benefit from some minimal documentation (doc strings)

src/watertap_contrib/reflo/costing/units/air_stripping.py
kurbansitterley: Done

src/watertap_contrib/reflo/property_models/air_water_equilibrium_properties.py
kurbansitterley: Done. Full documentation will be provided in readthedocs.

adam-a-a:
I guess the question is--do we want this to go before or after #91? We can (1) merge #91 then see what needs to be updated here and fix, (2) merge this and see what needs to be fixed in #91, or (3) try merging your branch for #91 into this branch and proactively start fixing anything that breaks due to this prop model; that way, we could proceed with merging #91 and immediately merge this afterwards (instead of waiting for #91 to get merged in first as suggested in (1)).

adam-a-a (approved):

kurbansitterley:
> I guess the question is--do we want this to go before or after #91? We can (1) merge #91 then see what needs to be updated here and fix, (2) merge this and see what needs to be fixed in #91, or (3) try merging your branch for #91 into this branch and proactively start fixing anything that breaks due to this prop model; that way, we could proceed with merging #91 and immediately merge this afterwards (instead of waiting for #91 to get merged in first as suggested in (1)).

I like (3), I will give it a try.

docs/technical_reference/unit_models/air_stripping.rst
```diff
@@ -0,0 +1,336 @@
+Air Stripping
+=============
+
+Air stripping uses towers packed with irregular shaped inert packing material
+to transfer volatile constituents from the liquid to the vapor phase.
+This air stripping model:
+   * supports steady-state only
+   * supports a single liquid phase only
+   * has a single user-specified target compound
+   * assumes isothermal and iosbaric conditions
```
adam-a-a, line 10:
```suggestion
   * assumes isothermal and isobaric conditions
```

docs/technical_reference/unit_models/air_stripping.rst
```diff
@@ -0,0 +1,336 @@
...
+* Feed air stream (``inlet``)
+* Effluent liquid stream (``outlet``)
+* Effluent air stream (``outlet``)
+
+A critical user input to the model is the target compound, specfied via the ``target`` keyword 
+in the unit model configuration. Removal of the target compound is determined by the mutable parameter
+``target_reduction_frac`` that has a default value of 0.9 (i.e., 90% removal).
+
+Sets
+----
```
adam-a-a, line 74: Did you intend to add sets here?

src/watertap_contrib/reflo/unit_models/air_stripping_0D.py
```diff
@@ -262,11 +263,11 @@ def build(self):
             doc="Factor to calculate pressure drop through tower",
         )
 
-        self.tower_height_safety_factor = Param(
+        self.tower_height_factor = Param(
             initialize=1.2,
             mutable=True,
             units=pyunits.dimensionless,
-            doc="Safety factor for tower height",
+            doc="Factor to calculate tower height",
```
adam-a-a, line 270: Minor, but any reason you moved away from calling a "safety factor"?

docs/technical_reference/unit_models/air_stripping.rst
kurbansitterley: oops just added

src/watertap_contrib/reflo/unit_models/zero_order/tests/test_chemical_softening.py
```diff
@@ -63,6 +63,7 @@ def chem_soft_frame(self):
             silica_removal=True,
             softening_procedure_type="excess_lime_soda",
         )
+        print(m.fs.soft.properties_in[0].is_property_constructed("flow_mol_phase_comp"))
```
zacharybinger, line 66: Can probably remove this now since not tested or make into an assert test

src/watertap_contrib/reflo/unit_models/air_stripping_0D.py
```diff
@@ -0,0 +1,1221 @@
...
+            doc="Diameter of tower access ports: 2-24 inches",
+        )
+
+        self.tower_pipe_diameter = Param(
+            initialize=6,
+            mutable=True,
+            units=pyunits.inches,
+            doc="Diameter of tower inlet and outlet piping: 2-24 inches",
+        )
+
```
zacharybinger, line 286: Since you state the diameters for tower_port and tower_pipe in the doc, should bounds be added that reflect that? bound=(2,24)

zacharybinger (approved):
Small comments, but overall approved

src/watertap_contrib/reflo/unit_models/air_stripping_0D.py
kurbansitterley: I would but you can't add bounds to a ``Param``

src/watertap_contrib/reflo/unit_models/zero_order/tests/test_chemical_softening.py
kurbansitterley: oops good catch

PR #89, opened 2023-12-19
Documentation for REFLO property models

kurbansitterley:
Adding documentation for:

- Basic water property package
- Air-water equilibrium property package

*Will remain a draft until #85 is merged

### DIFF
--- a/docs/technical_reference/property_models/air_water_equilibrium_properties.rst
+++ b/docs/technical_reference/property_models/air_water_equilibrium_properties.rst
@@ -113,7 +113,7 @@ Properties
    :header: "Description", "Symbol", "Variable", "Index", "Indexing Set", "Units"
 
    "Mass density of aqueous phase", ":math:`\rho`", "``dens_mass_phase``", "``[p]``", "``phase_list``", ":math:`\text{kg m}^{-3}`"
-   "Component molar flowrate", ":math:`N`", "``flow_mole_phase_comp``", "``[p, j]``", "``phase_component_set``", ":math:`\text{mol/s}`"
+   "Component molar flowrate", ":math:`N`", "``flow_mole_phase_comp``", "``[p, j]``", "``phase_component_set``", ":math:`\text{mol }\text{s}^{-1}`"
    "Component mass fraction", ":math:`x`", "``mass_frac_phase_comp``", "``[p, j]``", "``phase_component_set``", ":math:`\text{dimensionless}`"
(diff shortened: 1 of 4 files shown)


docs/technical_reference/property_models/air_water_equilibrium_properties.rst
```diff
@@ -0,0 +1,203 @@
...
+    * sets H2O as the solvent for the liquid phase;
+    * sets air as the solvent for the vapor phase;
+    * uses mass flowrate (kg/s), pressure, and temperature as state variables;
+    * does not support dynamics
+
+
+Configuration
+--------------
+
+The AWE property package has several configuration options depending on the users preferences.
```
adam-a-a, line 19:
```suggestion
The AWE property package has several configuration options depending on the user's preferences.
```

docs/technical_reference/property_models/basic_water_properties.rst
```diff
@@ -0,0 +1,95 @@
+.. _basic_water_prop_ref:
+
+Basic Water Property Package
+============================
+
+The basic water property package is meant to calculate the most basic of water properties for WaterTAP unit models
+from the volumetric flow rate and the component mass concentration.
+All components are assumed to be in the liquid phase and properties are indexed only to component where appropriate.
```
adam-a-a, line 8: My only thought is what differentiates this from the basic water property model in WaterTAP main? Is it only the usage of concentration and volumetric flow as state vars in place of component mass flows? Or are there other differences? No need to change anything now, but want to know whether we should create an issue to eventually consolidate the two.

kurbansitterley: Pretty much nothing. I anticipated getting more models from NMSU that would be able to make use of this prop pack, but that has not been the case (I thought they would be simpler). And using flow and concentration as state variables is more intuitive to me (as a non-ChemE). So I thought I was doing myself a favor.

The only model that uses this prop pack is chem softening, and in hindsight I should have just used one of the existing property packages. Now, using chemical softening in a flowsheet with any other property package requires translator blocks, so ended up making more work for us. What we should really do (not now) is convert that model to use a different property package.

kurbansitterley: Oh by "basic property model" do you mean the one used for ZO models? Answer is the same, just curious. 

It might be easier to just use that ZO property model for chem softening model. 

(Sidebar: we have two property packages that are imported as ``WaterParameterBlock`` in WaterTAP main - that might be something worth making an issue about)

kurbansitterley:
Merged with air stripping branch.

PR #98, opened 2024-01-11
Thermal unit models; SolarModelBase improvements

MuktaHardikar:
1. Added the flat plat collector physical model with some modifications- included ports for the heating fluid, calculating the heat_load. 
2. Added the thermal energy storage model. The calculation for the thermal_energy_capacity needs to be reviewed
3. Added the cost functions for both.


src/watertap_contrib/reflo/costing/solar/flat_plate.py
```diff
@@ -0,0 +1,151 @@
...
+        doc="Cost per area for solar collector",
+    )
+
+    blk.cost_per_volume_storage = pyo.Var(
+        initialize=2000,
+        units=costing.base_currency / pyo.units.m**3,
+        bounds=(0, None),
+        doc="Cost per volume for thermal storage",
+    )
+
```
Zhuoran29, line 39: Does the storage cost belong to the TES cost package? Or would it be double counted if the we also involve the TES component?

src/watertap_contrib/reflo/costing/solar/flat_plate.py
```diff
@@ -0,0 +1,151 @@
...
+        doc="Fixed operating cost of flat plate plant per kW capacity",
+    )
+
+    blk.variable_operating_by_generation = pyo.Var(
+        initialize=0,
+        units=costing.base_currency / (pyo.units.MWh * costing.base_period),
+        bounds=(0, None),
+        doc="Variable operating cost of flat plate plant per MWh generated",
+    )
+
```
Zhuoran29, line 67: I think this costing package uses the costing method from SAM, if so I guess maybe we also want to include the cost of balance of system and installation($1000 and $2500 in this case), or as a percentage of the equipment CAPEX, just to be consistent with the SAM default?

src/watertap_contrib/reflo/solar_models/surrogate/flat_plate/flat_plate_surrogate.py
```diff
@@ -0,0 +1,305 @@
...
+        )
+
+        self.storage_volume_constraint = Constraint(
+            expr=self.storage_volume
+            == pyunits.convert(
+                (
+                    (self.hours_storage * self.heat_load)
+                    / (
+                        self.specific_heat_water
+                        * self.temperature_cold
```
Zhuoran29, line 198: I might be wrong but I think we should use temp dif to calculate the energy available from the heat transfer fluid, as Mukta did in the TES model:
`* self.temperature_cold` --> `* (self.temp_design - self.temperature_cold)`

src/watertap_contrib/reflo/solar_models/zero_order/flat_plate_physical.py
```diff
@@ -0,0 +1,426 @@
...
+        @self.Constraint(doc="Heat generated annually")
+        def eq_heat_annual(b):
+            return b.heat_annual == b.heat * pyunits.convert(
+                1 * pyunits.year, to_units=pyunits.hour
+            )
+
+        # Need to check this
+        @self.Constraint(doc="Heat load of sysem")
+        def eq_heat_load(b):
+            return b.heat_load == b.collector_area_total * b.FR * b.ta * b.G_max * b.ta
```
Zhuoran29, line 362: Can also consider the heat loss component as in SAM and in the surrogate model, since the coefficent is already constructed, then
`b.heat_load == b.collector_area_total * b.FR * ( b.ta * b.G_max * b.ta - b.UL * b.rated_delta_T)`
where `b.rated_delta_T = 30 K` is the default temp difference for calculating rated capacity in SAM.

Zhuoran29:
It all looks great and useful!  Just have some details that I'm not sure about as commented in the code.

src/watertap_contrib/reflo/costing/solar/flat_plate.py
kurbansitterley: This is a good point. Surrogate has built in storage but physical does not. I will have to implement a type check to make this work.

src/watertap_contrib/reflo/costing/solar/flat_plate.py
MuktaHardikar: Kurby and I had a quick chat about this. It's not in the cost function because its a very small contribution  to the capital cost compared to other components. I might be wrong but I didn't see it included in the other cost functions either. But I added land area costs similar to the PV model.

src/watertap_contrib/reflo/solar_models/zero_order/flat_plate_physical.py
```diff
@@ -0,0 +1,475 @@
...
+
+        # ==========PARAMETERS==========
+
+        self.number_collectors = Param(
+            initialize=1,
+            units=pyunits.dimensionless,
+            doc="Number of collectors in array",
+        )
+
+        self.storage_volume = Param(
```
kurbansitterley, line 149: remove

src/watertap_contrib/reflo/solar_models/surrogate/flat_plate/data/training_flat_plate_surrogate.py
```diff
@@ -0,0 +1,333 @@
+import os
```
kurbansitterley, line 1: I'd like to build this functionality into the base class (i.e. running PySAM to generate data) but I want to merge this file in the meantime as a guide for how to do it for the flat plate surrogate since it is slightly different.

src/watertap_contrib/reflo/unit_models/tests/test_thermal_energy_storage.py
```diff
@@ -0,0 +1,176 @@
...
+        m.fs.tes.tes_hx_inlet.pressure.fix(101325)
+
+        m.fs.tes.tes_process_inlet.temperature.fix(55 + 273.15)
+        m.fs.tes.tes_process_inlet.flow_mass_phase_comp[0, "Liq", "H2O"].fix(0.05)
+        m.fs.tes.tes_process_inlet.flow_mass_phase_comp[0, "Vap", "H2O"].fix(0)
+        m.fs.tes.tes_process_inlet.pressure.fix(101325)
+
+        m.fs.tes.tes_hx_outlet.flow_mass_phase_comp[0, "Vap", "H2O"].fix(0)
+        m.fs.tes.tes_hx_outlet.pressure.fix(101325)
+        # m.fs.tes.initialize()
```
kurbansitterley, line 81:
```suggestion

```

kurbansitterley (approved):
LGTM - PV surrogate will need to be modified to conform to new SolarEnergyBase but that can be a separate PR.

src/watertap_contrib/reflo/core/solar_energy_base.py
```diff
@@ -56,16 +67,90 @@ class SolarEnergyBaseData(UnitModelBlockData):
         ),
     )
     CONFIG.declare(
-        "solar_energy_type",
+        "solar_model_type",
+        ConfigValue(
+            default=SolarModelType.surrogate,
+            domain=In(SolarModelType),
+            description="Solar model type construction flag",
+            doc="""Indicates what type of solar model will be constructed""",
```
adam-a-a, line 75: Should list the options here.

src/watertap_contrib/reflo/core/solar_energy_base.py
```diff
@@ -56,16 +67,90 @@ class SolarEnergyBaseData(UnitModelBlockData):
...
+    CONFIG.declare(
+        "surrogate_model_file",
         ConfigValue(
-            default="PV",
-            domain=In(SolarEnergyType),
-            description="Indicates type of solar energy source",
+            default=None,
+            domain=str,
+            description="Path to surrogate model file",
+            doc="""User provided surrogate model .json file. Must be in same directory as unit model file.""",
```
adam-a-a, line 84: > Must be in same directory as unit model file.

That seems like an undesirable requirement. Is there a reason for this restriction other than this was quickest and most convenient way to implement?

src/watertap_contrib/reflo/core/solar_energy_base.py
```diff
@@ -56,16 +67,90 @@ class SolarEnergyBaseData(UnitModelBlockData):
...
+            doc="""""",
+        ),
+    )
+    CONFIG.declare(
+        "training_fraction",
+        ConfigValue(
+            default=0.8,
+            domain=float,
+            description="Fraction of dataset to use as training data for surrogate",
+            doc="""""",
```
adam-a-a, line 147: Seems highly geared towards more of a `solar_energy_surrogate_base` rather than the possibility for a physical model, but then again, I haven't finished reading all the code yet.

src/watertap_contrib/reflo/core/solar_energy_base.py
```diff
@@ -111,3 +215,155 @@ def calculate_scaling_factors(self):
...
+            input_labels=self.input_labels,
+            output_labels=self.output_labels,
+            training_dataframe=self.data_training,
+        )
+
+        # Set PySMO options
+        # TODO: make this CONFIG for base?
+        self.trainer.config.basis_function = "gaussian"  # default = gaussian
+        self.trainer.config.solution_method = "algebraic"  # default = algebraic
+        self.trainer.config.regularization = True  # default = True
```
adam-a-a, line 239: For me, it makes more sense to include args for these in `create_rbf_surrogate`, but the more flexibility we enable, the more confident we should be in the model functionality/stability. So I guess it makes sense to keep as a TODO note for the time being (and maybe an issue if we don't have one yet?).

src/watertap_contrib/reflo/costing/solar/thermal_energy_storage.py
```diff
@@ -0,0 +1,118 @@
...
+        expr=blk.fixed_operating_cost
+        == tes_params.fixed_operating_by_capacity
+        * pyo.units.convert(tes.heat_load, to_units=pyo.units.kW)
+    )
+
+    # register the flows, e.g.,:
+    blk.costing_package.cost_flow(
+        tes.electricity,
+        "electricity",
+    )
```
adam-a-a, line 118: Any accounting for the possibility of auxiliary heat and associated costs?

src/watertap_contrib/reflo/solar_models/surrogate/flat_plate/data/pysam_run_flat_plate.py
```diff
@@ -0,0 +1,515 @@
...
+import time
+import multiprocessing
+from itertools import product
+import matplotlib.pyplot as plt
+import PySAM.Swh as swh
+from watertap_contrib.reflo.solar_models.surrogate.flat_plate.data.pyssc import (
+    PySSC,
+)
+
+
```
adam-a-a, line 15: More comments throughout this file would be helpful. As of now, I don't think we have as many comments as we might like.

src/watertap_contrib/reflo/solar_models/surrogate/flat_plate/data/pysam_run_flat_plate.py
```diff
@@ -0,0 +1,515 @@
+import json
```
adam-a-a, line 1: Add header?

src/watertap_contrib/reflo/solar_models/surrogate/flat_plate/data/pyssc.py
```diff
@@ -0,0 +1,534 @@
+import sys, os
```
adam-a-a, line 1: Add header?

src/watertap_contrib/reflo/solar_models/surrogate/flat_plate/data/pyssc.py
```diff
@@ -0,0 +1,534 @@
...
+    params["tech_model"] = TECH_NAME
+    params["financial_model"] = FINANCIAL_NAME
+    params["solar_resource_file"] = WEATHER_FILE
+    ssc = PySSC(params)
+    # results = ssc_sim_from_dict(ssc, params)
+    ssc.execute()
+    annual_Q_deliv = ssc.value("annual_Q_deliv")
+    P_pump = ssc.value("P_pump")
+
+    x = 1
```
adam-a-a, line 534:
```suggestion
    x = 1
```
Do we need this?

src/watertap_contrib/reflo/solar_models/surrogate/flat_plate/data/training_flat_plate_surrogate.py
```diff
@@ -0,0 +1,333 @@
...
+    print("Hot outlet temperature = {x:.1f} C".format(x=value(m.fs.temperature_hot)))
+    print("Annual heat output = {x:.2e} kWht".format(x=value(m.fs.heat_annual)))
+    print(
+        "Annual electricity input = {x:.2e} kWhe".format(
+            x=value(m.fs.electricity_annual)
+        )
+    )
+
+    x = 1
+    pass
```
adam-a-a, line 333: Do we need this?

src/watertap_contrib/reflo/solar_models/surrogate/flat_plate/data/untitled_cashloan.json
```diff
@@ -0,0 +1,135 @@
+{
```
adam-a-a, line 1: should filename be updated from "untitled_cashloan"?

src/watertap_contrib/reflo/unit_models/thermal_energy_storage.py
```diff
@@ -0,0 +1,454 @@
...
+
+        # From https://github.com/gmlc-dispatches/dispatches/blob/main/dispatches/properties/solarsalt_properties.py
+        # **** Requires Temperature in K
+        # Ref: (2015) Chang et al, Energy Procedia 69, 779 - 789
+        # Specific Heat Capacity as a function of Temperature, J/kg/K
+        # def specific_heat_cp(model):
+        #    return model.specific_heat_cp
+        #       == 1443 + 0.172 * model.temperature
+
+        # TODO: Add multiple tanks?
```
adam-a-a, line 208:
```suggestion
        # TODO: Add multiple tanks?
```
I would do that in a flowsheet rather than pack them in here (maybe?)-unless you could account for one lumped tank to account for multiple.

adam-a-a (approved):
I have a few comments but will approve so as not to hold up the PR. I will say this PR is pretty large, making it tougher or and a bit more confusing to review (thoroughly). For future PRs, would be nice to break these PRs up into much smaller chunks.

src/watertap_contrib/reflo/unit_models/thermal_energy_storage.py
```diff
@@ -0,0 +1,454 @@
...
+        self.add_inlet_port(name="tes_process_inlet", block=self.process_inlet_block)
+        self.add_outlet_port(name="tes_process_outlet", block=self.process_outlet_block)
+
+        # From https://github.com/gmlc-dispatches/dispatches/blob/main/dispatches/properties/solarsalt_properties.py
+        # **** Requires Temperature in K
+        # Ref: (2015) Chang et al, Energy Procedia 69, 779 - 789
+        # Specific Heat Capacity as a function of Temperature, J/kg/K
+        # def specific_heat_cp(model):
+        #    return model.specific_heat_cp
+        #       == 1443 + 0.172 * model.temperature
```
adam-a-a, line 206: Should delete commented code like this. Might want to include the ref to dispatches near the top of the class

src/watertap_contrib/reflo/solar_models/surrogate/flat_plate/data/pyssc.py
adam-a-a, line 1: Is this directory the best place to have this file? Shouldn't it be at a higher level that can later be utilized for flat plate and other units?

src/watertap_contrib/reflo/core/solar_energy_base.py
kurbansitterley: Good catch. No I don't think there is any reason this can't be anywhere if user is providing the whole path.

src/watertap_contrib/reflo/core/solar_energy_base.py
kurbansitterley: I think you are right and is something I realized too late in this implementation... we only have one physical solar model as of now and I believe the only thing this base class is doing for it is creating the ``heat`` and ``electricity`` variables.

src/watertap_contrib/reflo/core/solar_energy_base.py
kurbansitterley: Yes that is why I had it as a TODO for now. Get over the hurdle of this initial implementation and then refine.

src/watertap_contrib/reflo/costing/solar/thermal_energy_storage.py
kurbansitterley: That's a great question. @MuktaHardikar would like to hear your thoughts on this.

src/watertap_contrib/reflo/solar_models/surrogate/flat_plate/data/pyssc.py
kurbansitterley: This is sort of a custom implementation of PySAM as I understand it, not sure we should be putting a header on this one.

src/watertap_contrib/reflo/solar_models/surrogate/flat_plate/data/pyssc.py
kurbansitterley: Yeah I have no idea what this is here for. Most of these flat plate files were by someone who left our group.

kurbansitterley:
> I have a few comments but will approve so as not to hold up the PR. I will say this PR is pretty large, making it tougher or and a bit more confusing to review (thoroughly). For future PRs, would be nice to break these PRs up into much smaller chunks.

Agreed, this one got away from me as soon as I tried to add the flat plate surrogate to it.

I am going to address a few of these comments and then merge.

src/watertap_contrib/reflo/solar_models/surrogate/flat_plate/data/pyssc.py
kurbansitterley: Yes next step I think is to be able to generate data using REFLO and the SolarBaseClass... This would be moved to a core directory at that point is my initial though.

PR #103, opened 2024-01-31
Thermal unit models - Corrections

MuktaHardikar:
1. Updated the report function in the chemical softening model
2. In FPC: Changed the specific heat of heat transfer fluid to be a mutable parameter so it can be changed by user
3. In TES: Updated model to have water as heat transfer fluid and storage medium
4. In TES cost: Updated cost to be a function of storage volume and not thermal energy capacity (based on FPC surrogate/ SAM)
5. Updated the battery header


kurbansitterley (approved):
LGTM

src/watertap_contrib/reflo/unit_models/thermal_energy_storage.py
```diff
@@ -238,47 +183,48 @@ def build(self):
...
-        self.salt_mass = Var(
-            initialize=100,
-            units=pyunits.kg,
-            doc="Mass of salt in the thermal energy storage tank",
+        self.heat_transfer_fluid_density = Param(
+            initialize=1000,
+            units=pyunits.kg / pyunits.m**3,
+            mutable=True,
+            doc="Density of heat transfer fluid",
         )
```
adam-a-a, line 221: another example of something to put in a prop model

src/watertap_contrib/reflo/unit_models/zero_order/chemical_softening_zo.py
```diff
@@ -985,18 +985,18 @@ def _get_stream_table_contents(self, time_point=0):
...
         if (
             self.config.silica_removal
             and ["SiO2"] in self.config.property_package.solute_set
         ):
-            var_dict["MgCl2 dose"] = self.MgCl2_dosing()
+            expr_dict["MgCl2 dose"] = self.MgCl2_dosing
         else:
-            var_dict["MgCl2 dose"] = self.MgCl2_dosing.value
-        var_dict["Sludge produced"] = self.sludge_prod.value
+            var_dict["MgCl2 dose"] = self.MgCl2_dosing
```
adam-a-a, line 998: Best practice is to wrap in value(), e.g.: `value(self.my_var_or_expression)`. Using `.value` would work for a `Var` but not an `Expression`.

adam-a-a (approved):
LGTM

PR #104, opened 2024-02-02
Test surrogate

### LINKED ISSUE #95: Issues with trough surrogate test
Creating issue to track trough surrogate test failures.

I am sometimes getting tests passing locally and sometimes not. This may be as close as we can get to duplicating the issue 84. 

Here the tests passed and then I separated the large and small tests into separate classes (didn't modify tests) and ran the tests again and didn't get test failures, but some other failure that caused pytest to stop and was indicated by the ``%`` symbol at the end of the pytest output.

Screenshot 2023-12-22 at 4 39 41 PM

dguittet:
Fixed the random state for the surrogate tests and now they are passing. The tests still include the PySMO surrogate fitting. I re-ran the tests on Github CI a few times and it passed each time, so I hope the flakiness has been resolved.

`src/watertap_contrib/reflo/solar_models/surrogate/trough/data/training_trough_surrogate.py` is a script that runs the training. It's currently broken due to some changes to TroughSurrogateData. What should be done with this file? If the surrogate training functionality is now in another part of the code, then perhaps this file could be deleted? Otherwise, I could fix this file but there should probably be some way to check that it doesn't keep breaking in the future.

Fixes #95 via approach (1)


src/watertap_contrib/reflo/solar_models/surrogate/trough/test_trough_surrogate.py
```diff
@@ -46,6 +47,8 @@
     unscaled_variables_generator,
 )
 
+random.seed(42)
```
adam-a-a, line 50: Really excited to see that you got this approach to work! One question here--how does setting this random seed get applied in the test code below?

adam-a-a:
> `src/watertap_contrib/reflo/solar_models/surrogate/trough/data/training_trough_surrogate.py` is a script that runs the training. It's currently broken due to some changes to TroughSurrogateData. What should be done with this file? If the surrogate training functionality is now in another part of the code, then perhaps this file could be deleted? Otherwise, I could fix this file but there should probably be some way to check that it doesn't keep breaking in the future.

Still need to reflect on this a bit, but the PR looks great to me!

src/watertap_contrib/reflo/solar_models/surrogate/trough/test_trough_surrogate.py
dguittet: It was to catch any other code that calls random without setting a seed. I'm not sure if there are such calls being made, so I've removed this line to see if the tests still pass

kurbansitterley:
> > `src/watertap_contrib/reflo/solar_models/surrogate/trough/data/training_trough_surrogate.py` is a script that runs the training. It's currently broken due to some changes to TroughSurrogateData. What should be done with this file? If the surrogate training functionality is now in another part of the code, then perhaps this file could be deleted? Otherwise, I could fix this file but there should probably be some way to check that it doesn't keep breaking in the future.
> 
> Still need to reflect on this a bit, but the PR looks great to me!

When we first started making surrogates, I kept all of the files used to make the surrogates around so they didn't disappear somewhere, including this one. Now that we have a grip on surrogate building, we can safely delete any files not needed to run the tests. Can either do it in this PR or another.

src/watertap_contrib/reflo/core/solar_energy_base.py
```diff
@@ -301,7 +301,9 @@ def get_surrogate_data(
                 self.pickle_df = self.pickle_df[
                     (self.pickle_df[col] >= lo) & (self.pickle_df[col] <= hi)
                 ].copy()
-        self.data = self.pickle_df.sample(n=self.config.number_samples)
+        self.data = self.pickle_df.sample(
+            n=self.config.number_samples, random_state=len(self.pickle_df)
```
kurbansitterley, line 305: What does including the ``random_state`` argument do here? Is this part of ensuring we are generating a consistent surrogate?

kurbansitterley (approved):
LGTM, seems straightforward enough. Thanks for finishing this!

src/watertap_contrib/reflo/core/solar_energy_base.py
dguittet: Thanks for the quick review! My guess is this particular sample selection doesn't result in issues in the surrogate process.

PR #105, opened 2024-02-05
replace PySSC with PySAM for Flat Plate PV data

### LINKED ISSUE #99: Generating data for flat plate surrogate only works for Windows
_As of writing the flat plate surrogate has not been merged_

To get a good surrogate for the flat plate collector, modifications had to be made to the SAM model (or perhaps more accurately as I understand it- @Matthew-Boyd  had to create a custom flat plate model). As it stands now this will only run on Windows machines. So as we need to create new surrogates for different location, or different range of inputs, etc. it can only be done on a Windows machine.

This is way over my head, but running on MacOS the ``pysam_run_flat_plate.py`` is looking for ``ssc.dylib`` and we only have the ``ssc.dll``.

dguittet:
For `pysam_run_flat_plate.py`, the current data generation uses a custom SSC dynamic library saved in the `data` folder due to new features that had not yet been released in PySAM.

This PR updates the data generation to use the newest PySAM version, and adds a test of running the flat plate PV model ("Solar Water Heating" in SAM) to use as a regression test to make sure the `heat_annual` and `electricity_annual` are the same.

Fixes #99


kurbansitterley (approved):
LGTM, thanks for tackling this.

kurbansitterley:
> For `pysam_run_flat_plate.py`, the current data generation uses a custom SSC dynamic library saved in the `data` folder due to new features that had not yet been released in PySAM.
> 
> This PR updates the data generation to use the newest PySAM version, and adds a test of running the flat plate PV model ("Solar Water Heating" in SAM) to use as a regression test to make sure the `heat_annual` and `electricity_annual` are the same. This is currently a new test file, but I could see this test being merged into `test_flat_plate_surrogate.py` instead of being a whole new file.
> 
> Fixes #99

So this implements a way to generate data for the flat plate surrogate? Could it be generalized to generate data for any input range and location?

dguittet:
> > For `pysam_run_flat_plate.py`, the current data generation uses a custom SSC dynamic library saved in the `data` folder due to new features that had not yet been released in PySAM.
> > This PR updates the data generation to use the newest PySAM version, and adds a test of running the flat plate PV model ("Solar Water Heating" in SAM) to use as a regression test to make sure the `heat_annual` and `electricity_annual` are the same. This is currently a new test file, but I could see this test being merged into `test_flat_plate_surrogate.py` instead of being a whole new file.
> > Fixes #99
> 
> So this implements a way to generate data for the flat plate surrogate? Could it be generalized to generate data for any input range and location?

The original method was using a custom SSC binary stored in `data`. This modification now uses that SSC binary that comes with the PySAM downloaded as part of the requirements. The file can definitely be modified to change the input range and location.

mjprilliman (approved):

PR #113, opened 2024-06-16
Flowsheet development of batch-VAGMD simulation

Zhuoran29:
This PR constructs the multiperiod model of VAGMD-batch system directly in a flowsheet, which enables the dynamic connection with upstream and downstream treatment processes. 

The major update and difference from the previous model are:

- The previous model was wrapped in a unit model with an assumed recovery ratio that determines the required number of periods in the multiperiod model to meet the target. 
- This flowsheet takes in the number of periods as an input, and calculate the corresponding system performance (e.g. overall recovery ratio)
- Updated the filename of the previous unit model to: `VAGMD_batch_multiperiod_unit_model.py`
- A sample flowsheet is built and executed using this new structure in: `VAGMD_batch_multiperiod_flowsheet.py`


kurbansitterley (approved):
I modified the test file a bit, but other than that it LGTM! Nice job @Zhuoran29

src/watertap_contrib/reflo/analysis/multiperiod/vagmd_batch/VAGMD_batch_multiperiod_flowsheet.py
```diff
@@ -0,0 +1,495 @@
...
+        "cooling_system_type": cooling_system_type,
+        "cooling_inlet_temp": cooling_inlet_temp,
+    }
+
+    mp.build_multi_period_model(
+        model_data_kwargs={t: model_options for t in range(n_time_points)},
+        flowsheet_options=model_options,
+        initialization_options={
+            "feed_flow_rate": feed_flow_rate,
+            "feed_salinity": feed_salinity,
```
MuktaHardikar, line 183: Feed salinity and flow rates are not fixed in the fix_dof_and_initialize function currently. Do we still need this here in that case?

kurbansitterley: I'd say this is a question for @Zhuoran29

kurbansitterley: though @MuktaHardikar I see that tests were passing before and after your change in f383125 so wondering why you wanted to release state?

Zhuoran29: @MuktaHardikar  No, those two redundant arguments were never used in the initialization routine. Now I removed them in all associated functions. Thanks!

MuktaHardikar (approved):
Thanks for clarifying my comments. Everything LGTM! Great work Zhuoran!  In the future having ports in the multiperiod flowsheet unit model would be helpful but not urgent.

PR #114, opened 2024-06-24
Forward osmosis unit model and Trevi's system flowsheet

Zhuoran29:
This PR integrates the forward osmosis (FO) model developed from the Trevi System's pilot project and includes the following features:

- A property package describing the proprietary draw solution properties used by Trevi
- An FO unit model and its associated test file
- An example flowsheet of building Trevi's FO configuration
- A cost model based upon the Trevi's system costing numbers
- Documentation of both the unit model and example flowsheet


kurbansitterley:
@Zhuoran29 can you check the failing tests? Just updated the repo to use WaterTAP v 1.0 so could be related to that.

src/watertap_contrib/reflo/analysis/example_flowsheets/test/test_fo_trevi_flowsheet.py
```diff
@@ -0,0 +1,129 @@
...
+)
+
+from watertap_contrib.reflo.analysis.example_flowsheets.fo_trevi_flowsheet import (
+    build_fo_trevi_flowsheet,
+    fix_dof_and_initialize,
+    get_flowsheet_performance,
+)
+from watertap_contrib.reflo.costing import TreatmentCosting
+from idaes.core import UnitModelCostingBlock
+from idaes.core.solvers import get_solver
```
kurbansitterley, line 27:
```suggestion
from watertap.core.solvers import get_solver
```

src/watertap_contrib/reflo/property_models/fo_draw_solution_properties.py
```diff
@@ -0,0 +1,812 @@
...
+import idaes.logger as idaeslog
+import idaes.core.util.scaling as iscale
+
+from idaes.core.util.exceptions import (
+    ConfigurationError,
+    InitializationError,
+    PropertyPackageError,
+)
+
+from pyomo.environ import (
```
kurbansitterley, line 48: import order should be

other pkg imports (e.g., pandas, numpy)
pyomo imports
idaes imports
watertap imports
watertap-reflo imports

src/watertap_contrib/reflo/property_models/fo_draw_solution_properties.py
```diff
@@ -0,0 +1,812 @@
...
+    MaterialBalanceType,
+    MaterialFlowBasis,
+    PhysicalParameterBlock,
+    StateBlock,
+    StateBlockData,
+    declare_process_block_class,
+)
+from idaes.core.base.components import Solvent, Solute
+from idaes.core.base.phases import LiquidPhase
+from idaes.core.solvers.get_solver import get_solver
```
kurbansitterley, line 28: use watertap solver

``from watertap.core.solvers import get_solver``

src/watertap_contrib/reflo/property_models/tests/test_FO_draw_solution_properties.py
```diff
@@ -0,0 +1,217 @@
...
+    Suffix,
+    value,
+    Var,
+    Set,
+    assert_optimal_termination,
+)
+from pyomo.util.check_units import assert_units_consistent
+from idaes.core import FlowsheetBlock
+
+from idaes.core.solvers.get_solver import get_solver
```
kurbansitterley, line 26: use watertap solver

src/watertap_contrib/reflo/unit_models/zero_order/forward_osmosis_zo.py
```diff
@@ -0,0 +1,750 @@
...
+# Import IDAES cores
+from idaes.core import (
+    declare_process_block_class,
+    UnitModelBlockData,
+    useDefault,
+)
+from idaes.core.util.model_statistics import degrees_of_freedom
+from idaes.core.util.config import is_physical_parameter_block
+import idaes.core.util.scaling as iscale
+from idaes.core.solvers import get_solver
```
kurbansitterley, line 33: use watertap solver

kurbansitterley (changes requested):
haven't reviewed all the content yet, but left a few comments to fix before then.

Zhuoran29:
> @Zhuoran29 can you check the failing tests? Just updated the repo to use WaterTAP v 1.0 so could be related to that.

@kurbansitterley , Sorry for the late response. Yes, I've been looking into it and it seemed to need some more efforts to solve the convergence issue (seems related to an `expr_if`-like expression, even when scaling factor and jacobian matrix both look good). I'll try to fix it asap.

kurbansitterley:
> > @Zhuoran29 can you check the failing tests? Just updated the repo to use WaterTAP v 1.0 so could be related to that.
> 
> @kurbansitterley , Sorry for the late response. Yes, I've been looking into it and it seemed to need some more efforts to solve the convergence issue (seems related to an `expr_if`-like expression, even when scaling factor and jacobian matrix both look good). I'll try to fix it asap.

If it is related to the ``expr_if``, you should do a totally fresh install of your environment. There was a compatibility issue with a previous version of Pyomo that has since been fixed. So I'd recommend checking if that solves your problem first before going too much further down the rabbit hole.

Zhuoran29:
> > > @Zhuoran29 can you check the failing tests? Just updated the repo to use WaterTAP v 1.0 so could be related to that.
> > 
> > 
> > @kurbansitterley , Sorry for the late response. Yes, I've been looking into it and it seemed to need some more efforts to solve the convergence issue (seems related to an `expr_if`-like expression, even when scaling factor and jacobian matrix both look good). I'll try to fix it asap.
> 
> If it is related to the `expr_if`, you should do a totally fresh install of your environment. There was a compatibility issue with a previous version of Pyomo that has since been fixed. So I'd recommend checking if that solves your problem first before going too much further down the rabbit hole.

Zhuoran29:
Hey @kurbansitterley, just fixed the converging issue of this PR. Sorry for taking long.

src/watertap_contrib/reflo/unit_models/zero_order/tests/test_forward_osmosis.py
```diff
@@ -0,0 +1,272 @@
...
+from watertap_contrib.reflo.costing import REFLOCosting
+
+# -----------------------------------------------------------------------------
+# Get default solver for testing
+solver = get_solver()
+
+
+class TestFO:
+    @pytest.fixture(scope="class")
+    def FO_frame(self):
```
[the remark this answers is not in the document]
Zhuoran29: Yes, this case is the one that we originally received from Trevi in an Excel model.

src/watertap_contrib/reflo/unit_models/zero_order/forward_osmosis_zo.py
```diff
@@ -0,0 +1,886 @@
...
+        var_dict["Membrane temperature (K)"] = self.membrane_temp
+
+        return {"vars": var_dict}
+
+    @property
+    def default_costing_method(self):
+        return cost_forward_osmosis
+
+
+if __name__ == "__main__":
```
kurbansitterley, line 748: This is interesting, we typically don't have a __main__ function in separate unit models. I am not sure if this is a good idea from a software-package-development perspective. My gut is to remove this but interested if @adam-a-a has any thoughts. I am also interested if @Zhuoran29 included this for a specific reason?

src/watertap_contrib/reflo/unit_models/zero_order/forward_osmosis_zo.py
```diff
@@ -0,0 +1,886 @@
...
+
+        # Initial guess of weak draw properties
+        state_args_weak_draw = deepcopy(state_args_draw_solution)
+        for p, j in blk.weak_draw_props.phase_component_set:
+            if p == "Liq" and j == "H2O":
+                state_args_weak_draw["flow_mass_phase_comp"][(p, j)] = (
+                    state_args["flow_mass_phase_comp"][(p, j)]
+                    * blk.recovery_ratio
+                    / blk.nanofiltration_recovery_ratio
+                    * 1.5
```
kurbansitterley, line 510: What is this number? The initialization routine looks good and I liked that you called out what some of these other constants in the routine are for.

src/watertap_contrib/reflo/unit_models/zero_order/forward_osmosis_zo.py
```diff
@@ -0,0 +1,886 @@
...
+        state_args_strong_draw = deepcopy(state_args_draw_solution)
+        for p, j in blk.strong_draw_props.phase_component_set:
+            if p == "Liq" and j == "DrawSolution":
+                state_args_strong_draw["flow_mass_phase_comp"][(p, j)] = (
+                    state_args_weak_draw["flow_mass_phase_comp"][(p, "H2O")]
+                )
+            elif p == "Liq" and j == "H2O":
+                state_args_strong_draw["flow_mass_phase_comp"][(p, j)] = (
+                    state_args_strong_draw["flow_mass_phase_comp"][(p, "DrawSolution")]
+                    * 0.25  # typical draw : water in strong draw solution
```
kurbansitterley, line 536: Like here- add the description of constant to the 1.5 above.

src/watertap_contrib/reflo/unit_models/zero_order/forward_osmosis_zo.py
```diff
@@ -0,0 +1,886 @@
...
+                    / pyunits.s
+                )
+
+        blk.brine_props.initialize(
+            outlvl=outlvl,
+            optarg=optarg,
+            solver=solver,
+            state_args=state_args_brine,
+        )
+
```
kurbansitterley, line 485:
```suggestion
init_log.info("Initialization of briner properties {}.".format(idaeslog.condition(res)))
```

Since there are so many state blocks in this model, it would be useful to add log messages for each of them imo.

src/watertap_contrib/reflo/unit_models/zero_order/tests/test_forward_osmosis.py
```diff
@@ -0,0 +1,272 @@
...
+from watertap.core.solvers import get_solver
+from watertap.property_models.seawater_prop_pack import SeawaterParameterBlock
+from watertap_contrib.reflo.unit_models.zero_order.forward_osmosis_zo import (
+    ForwardOsmosisZO,
+)
+from watertap_contrib.reflo.property_models.fo_draw_solution_properties import (
+    FODrawSolutionParameterBlock,
+)
+from watertap_contrib.reflo.costing import REFLOCosting
+
```
kurbansitterley, line 52: I see there is a costing test in the Trevi flowsheet test, so I am ok with there not being one here.

src/watertap_contrib/reflo/analysis/example_flowsheets/test/test_fo_trevi_flowsheet.py
```diff
@@ -0,0 +1,132 @@
...
+        m.fs.costing = TreatmentCosting()
+        m.fs.costing.base_currency = pyunits.USD_2020
+
+        # Create cost block for FO
+        m.fs.fo.costing = UnitModelCostingBlock(flowsheet_costing_block=m.fs.costing)
+
+        # Add LCOW component
+        m.fs.costing.cost_process()
+        m.fs.costing.add_annual_water_production(m.fs.system_capacity)
+        m.fs.costing.add_LCOW(m.fs.system_capacity)
```
kurbansitterley, line 98: Can you add a check that ```m.fs.system_capacity``` equals however this flow is being represented in the unit model (i.e. ```product_prop.flow_vol_phase["Liq"]``` or whatever.)

src/watertap_contrib/reflo/costing/units/forward_osmosis_zo.py
```diff
@@ -0,0 +1,153 @@
...
+        bounds=(0, None),
+        doc="Specific electric energy consumption (kWh/m3)",
+    )
+
+    blk.cost_chemicals = pyo.Var(
+        initialize=0.07,
+        units=costing.base_currency / pyo.units.m**3,
+        bounds=(0, None),
+        doc="Cost of chemicals per m3 product",
+    )
```
kurbansitterley, line 72: Same question as above.

src/watertap_contrib/reflo/costing/units/forward_osmosis_zo.py
```diff
@@ -0,0 +1,153 @@
...
+import pyomo.environ as pyo
+import idaes.core.util.scaling as iscale
+from watertap.costing.util import register_costing_parameter_block
+from watertap_contrib.reflo.costing.util import (
+    make_capital_cost_var,
+    make_fixed_operating_cost_var,
+)
+
+# Costing numbers are from data provided by Trevi (2021),
+# and should be applied to a Trevi FO flowsheet.
```
kurbansitterley, line 22: Just saw this comment, ignore my comments below. Though, if Trevi provided this in 2021, should all the costing parameters be in ```pyunits.USD_2021``` rather than ```base_currency```?

src/watertap_contrib/reflo/costing/units/forward_osmosis_zo.py
```diff
@@ -0,0 +1,153 @@
...
+    test_fs = fo.flowsheet()
+
+    brine = fo.brine_props[0]
+
+    blk.costing_package.add_cost_factor(blk, None)
+    blk.annual_dist_production = pyo.units.convert(
+        test_fs.system_capacity, to_units=pyo.units.m**3 / pyo.units.year
+    )
+    blk.capital_cost_constraint = pyo.Constraint(
+        expr=blk.capital_cost
```
kurbansitterley, line 116: If all costing params are indeed for ```pyunits.USD_2021```, you need to convert all costing variables to ```base_currency``` (and ideally ```base_period``` though this is always year to this point, it is not clear to me if defining the ```base_period``` to be something besides year like the duration of a multiperiod flowsheet would be useful for costing multiperiod flowsheets on an annual basis).

src/watertap_contrib/reflo/costing/units/forward_osmosis_zo.py
```diff
@@ -0,0 +1,153 @@
...
+    parameter_block_name="forward_osmosis",
+)
+def cost_forward_osmosis(blk):
+
+    fo_params = blk.costing_package.forward_osmosis
+    make_capital_cost_var(blk)
+    make_fixed_operating_cost_var(blk)
+
+    fo = blk.unit_model
+    test_fs = fo.flowsheet()
```
kurbansitterley, line 107: could be ```fo_fs``` unless ```test_``` is there for a reason?

src/watertap_contrib/reflo/costing/units/forward_osmosis_zo.py
```diff
@@ -0,0 +1,153 @@
...
+import idaes.core.util.scaling as iscale
+from watertap.costing.util import register_costing_parameter_block
+from watertap_contrib.reflo.costing.util import (
+    make_capital_cost_var,
+    make_fixed_operating_cost_var,
+)
+
+# Costing numbers are from data provided by Trevi (2021),
+# and should be applied to a Trevi FO flowsheet.
+# Mass and energy flows should be calculated from a Trevi FO flowsheet.
```
kurbansitterley, line 23: Does this mean that this costing package is not compatible with a standalone FO flowsheet?

src/watertap_contrib/reflo/property_models/fo_draw_solution_properties.py
```diff
@@ -0,0 +1,815 @@
...
+#
+# Please see the files COPYRIGHT.md and LICENSE.md for full copyright and license
+# information, respectively. These files are also available online at the URL
+# "https://github.com/watertap-org/watertap/"
+#################################################################################
+
+"""
+This module contains a basic property package for simple water treatment models.
+Volumetric flow and component concentration are used to determine mass flow. 
+"""
```
kurbansitterley, line 16: Update this comment (I don't think it is accurate).

src/watertap_contrib/reflo/property_models/fo_draw_solution_properties.py
```diff
@@ -0,0 +1,815 @@
...
+                "heat_separation_phase": {"method": "_heat_separation_phase"},
+                "enth_flow": {"method": "_enth_flow"},
+            }
+        )
+
+        obj.define_custom_properties(
+            {
+                # "heat_separation_phase": {"method": "_heat_separation_phase"},
+            }
+        )
```
kurbansitterley, line 211: Can be removed?

src/watertap_contrib/reflo/property_models/fo_draw_solution_properties.py
```diff
@@ -0,0 +1,815 @@
...
+            initialize=1e6,
+            bounds=(-1e10, 1e10),
+            units=pyunits.J / pyunits.s,
+            doc="Heat of liquid separation",
+        )
+        self.liquid_separation = Var(
+            initialize=0.1,
+            units=pyunits.dimensionless,
+            doc="Indication of whether liquid separation will happen",
+        )
```
kurbansitterley, line 631: I see why this may have been removed from the ```define_custom_properties``` dict above, since it looks like this is meant to be some kind of boolean indicator and I imagine that is tricky to deal with in a property package. I saw you had a ```Binary``` import and probably discovered that IPOPT can't handle it (i.e., it is treated as a float), but if you would like to include this and need this functionality, you can try ```Disjunct```, otherwise might want to remove this entire ```_heat_separation_phase```.

kurbansitterley (changes requested):
Overall I am very impressed with this huge PR and think it is in great shape. There are a few code things to address and comments I left.

src/watertap_contrib/reflo/unit_models/zero_order/forward_osmosis_zo.py
Zhuoran29: Thanks, it's supposed to be fully removed.

src/watertap_contrib/reflo/unit_models/zero_order/forward_osmosis_zo.py
Zhuoran29: The comment is added now.

src/watertap_contrib/reflo/property_models/tests/test_FO_draw_solution_properties.py
Zhuoran29: Updated to watertap solver.

src/watertap_contrib/reflo/unit_models/zero_order/forward_osmosis_zo.py
Zhuoran29: Updated to watertap solver.

src/watertap_contrib/reflo/property_models/fo_draw_solution_properties.py
Zhuoran29: Updated to watertap solver.

src/watertap_contrib/reflo/property_models/fo_draw_solution_properties.py
Zhuoran29: Order adjusted.

src/watertap_contrib/reflo/unit_models/zero_order/forward_osmosis_zo.py
Zhuoran29: The messages are already scripted in the property package and will display if we set `outlvl = idaeslog.INFO_HIGH` (currently is `idaeslog.NOTSET`). I'll leave it as is unless you think we should change the significance of the log messages.

src/watertap_contrib/reflo/costing/units/forward_osmosis_zo.py
```diff
@@ -0,0 +1,153 @@
...
+        doc="Base price of the labor cost ($/m3)",
+    )
+
+    blk.labor_cost_index = pyo.Var(
+        initialize=-0.178,
+        units=pyo.units.dimensionless,
+        bounds=(None, 0),
+        doc="Scaling factor of Trevi FO system capital cost",
+    )
+
```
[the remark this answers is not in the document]
Zhuoran29: It's not particularly high (like 0.07 to 0.15 cents/m3) so I removed it from the cost package, together with chemical cost and insurance.

src/watertap_contrib/reflo/costing/units/forward_osmosis_zo.py
Zhuoran29: Chemical cost is removed.

src/watertap_contrib/reflo/costing/units/forward_osmosis_zo.py
Zhuoran29: Yes, I changed the units to `USD_2021`.

src/watertap_contrib/reflo/costing/units/forward_osmosis_zo.py
Zhuoran29: For steady-state operation, I think converting to `base_period == pyunits.year` or other units should not impact anything. If it's variable operation, I guess it might overestimate the cost if using year within an hourly/daily multiperiod fs, so I'd use `base_currency / flow_rate` for those cases.
For this model, the variables are in `pyunits.USD_2021` now and I assume the solver will just convert those values to whatever the `base_currency` is?

src/watertap_contrib/reflo/costing/units/forward_osmosis_zo.py
Zhuoran29: Yes, it should be fo_fs and is corrected.

src/watertap_contrib/reflo/costing/units/forward_osmosis_zo.py
Zhuoran29: Yes. The costing package only applies to the whole flowsheet and not the unit model, so the heat exchangers, pumps, etc. are not separately considered in the flowsheet.

src/watertap_contrib/reflo/property_models/fo_draw_solution_properties.py
Zhuoran29: Thanks, updated!

src/watertap_contrib/reflo/property_models/fo_draw_solution_properties.py
```diff
@@ -0,0 +1,815 @@
...
+
+__author__ = "Zhuoran Zhang"
+
+# Set up logger
+_log = idaeslog.getLogger(__name__)
+
+
+@declare_process_block_class("FODrawSolutionParameterBlock")
+class FODrawSolutionParameterBlockData(PhysicalParameterBlock):
+    """
```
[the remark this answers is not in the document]
Zhuoran29: Updated.

src/watertap_contrib/reflo/property_models/fo_draw_solution_properties.py
Zhuoran29: Yes, it should be removed, as this property is already in the "normal" property list above it.

src/watertap_contrib/reflo/property_models/fo_draw_solution_properties.py
```diff
@@ -0,0 +1,812 @@
...
+        self.eq_enth_mass_phase = Constraint(
+            self.params.phase_list, rule=rule_enth_mass_phase
+        )
+
+    def _enth_flow(self):
+        # enthalpy flow expression for get_enthalpy_flow_terms method
+
+        def rule_enth_flow(b):  # enthalpy flow [J/s]
+            return (
+                sum(b.flow_mass_phase_comp["Liq", j] for j in b.params.component_list)
```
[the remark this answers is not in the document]
Zhuoran29: Yes, as mentioned above it's already in the property list and now the custom property dict is deleted.

src/watertap_contrib/reflo/analysis/example_flowsheets/test/test_fo_trevi_flowsheet.py
Zhuoran29: Sure, check is now added.

Zhuoran29:
@kurbansitterley , thanks for the constructive suggestions. I've modified some of them and responded to those that I didn't change. Please feel free to follow with further comments.

src/watertap_contrib/reflo/costing/units/forward_osmosis_zo.py
kurbansitterley: Yes correct. We always want to make sure we are converting our costing output to the ```base_currency``` set at the flowsheet level.

src/watertap_contrib/reflo/property_models/fo_draw_solution_properties.py
kurbansitterley: @Zhuoran29 curious if you ever made ust of the ```Disjunct``` for this model/flowsheet? If so, consider converting this PR to a draft until you are finished with the feature.

Zhuoran29: @kurbansitterley I gave it a try but it didn't work out. I only found a few examples using pyomo disjunct and those examples didn't work themselves on my machine. 
The current approach of using `Expr_if` is working fine, should we stick to this approach before I educate myself enough about this feature?

kurbansitterley: That's ok, I am not very familiar with the ```Disjunct``` but just wanted to check if you were using it before we merged this PR. 

If the situation can be handled with an if statement, and ```Expr_if``` is working, I don't see any reason to force it. I imagine the ```Disjunct``` would be needed if this decision were "optimizable," but I don't think it is.

Zhuoran29: Yes, `Disjunct` is a more generalized feature than `Expr_if` and is more applicable when optimizable decision is involved. In this flowsheet the indicator/decision is fixed based on the configuration and shouldn't be mutable.

src/watertap_contrib/reflo/costing/units/forward_osmosis_zo.py
kurbansitterley: Sorry @Zhuoran29, I think I skimmed/misunderstood your question. 

The solver doesn't do any unit conversions. The ```FlowsheetBlockCosting``` IDAES class the WaterTAP costing pkg inherits (which we then inherit) will convert all aggregated costs to ```base_currency``` and ```base_period``` units. So while it is true that we don't need to make the conversion here for the units to be correct on the flowsheet level, if we report costing at the unit process level (which seems likely) we will have to then convert all those values to the ```base_currency```. This is something that I typically forget to do (and check for when reviewing), but it has burned me in the past.

Zhuoran29: Just remembered to do this: now `base_period` and `base_currency` are applied to the costing variables. Thanks for the clarification!

kurbansitterley (approved):
Great job @Zhuoran29

PR #119, opened 2024-07-08
Add back costing for MgCl2 flow in Chem Softening

kurbansitterley:
There was an issue with the ``Expr_if`` in Pyomo that was causing errors for costing the MgCl2 flow that depends on use of ``Expr_if``. 

This issue has been resolved and so this PR just adds back costing of the MgCl2 flow and adds a test to check that it is working.


MuktaHardikar:
Updated the MgCl2 cost and checked the MgCl2 dosing value.

MuktaHardikar (approved):

PR #121, opened 2024-07-18
Multi-effect crystallizer unit model

Zhuoran29:
This PR adds:
- ```CrystallizerEffect``` unit model
- ```MultiEffectCrystallizer``` unit model that is comprised of _n_ ```CrystallizerEffect``` unit models connected via constraints
- test files for both
- costing for ```MultiEffectCrystallizer``` and ```CrystallizerEffect```
- **importantly, this PR also points to ```watertap==1.0.0rc0``` in setup.py**

TODO:

- [x] check units on heat transfer equations
- [x] test file for ```CrystallizerEffect``` 
- [x] determine if ```flow_vol_phase['Vap']``` is the proper flow to cost for steam
- [x] remove __name__ == "__main__" in unit model files


kurbansitterley:
@Zhuoran29 can you run black so we can see the results of the GH tests please?

Zhuoran29:
> @Zhuoran29 can you run black so we can see the results of the GH tests please?

Sure, it seems the cost package is not adjusted well and I'll look into it.

src/watertap_contrib/reflo/unit_models/zero_order/tests/test_crystallizer_watertap.py
```diff
@@ -0,0 +1,787 @@
...
+
+        for _ in badly_scaled_var_generator(m):
+            assert False
+
+    @pytest.mark.requires_idaes_solver
+    @pytest.mark.component
+    def test_initialize(self, Crystallizer_frame):
+        # Add costing function, then initialize
+        m = Crystallizer_frame
+        # m.fs.costing = WaterTAPCosting()
```
adam-a-a, line 334: @Zhuoran29 just skimming this from my phone, but maybe the failed tests are due to the costing lines being commented out here.

adam-a-a: In any case, I’d recommend just having a separate test_costing test rather than bundle it up in the initialization test.

Zhuoran29:
I added the costing package from the WaterTAP repo and a test file for the flowsheet, where both simulation and optimization modes were tested.

src/watertap_contrib/reflo/unit_models/zero_order/tests/test_crystallizer_watertap.py
Zhuoran29: Now the costing package is rebuilt in REFLO and tested separately.

Zhuoran29:
Hey @kurbansitterley @adam-a-a , forgot to tag you when I updated this PR.

kurbansitterley:
I have just noticed both watertap-org/watertap#1482 and watertap-org/watertap#1487 that are modifying either the unit model itself or the property package it uses. If we are going to inherit ```Crystallization``` for the ```MultiEffectCrystallizer```, we might want to wait to see what changes are made there, especially if we want to keep updating the watertap dependency.

Zhuoran29 (commented):
All looks good! Just two comments regarding the change of unit.

src/watertap_contrib/reflo/costing/units/multi_effect_crystallizer.py
```diff
@@ -0,0 +1,381 @@
...
+    if effect_number == 1:
+        effect.steam_flow = pyo.units.convert(
+            (effect.work_mechanical[0] / _compute_steam_specific_energy(effect)),
+            to_units=pyo.units.m**3 / pyo.units.s,
+        )
+        costing_package.cost_flow(
+            effect.steam_flow,
+            "steam",
+        )
+        # costing_package.cost_flow(effect.heating_steam[0].flow_vol_phase["Vap"],
```
kurbansitterley, line 299: @Zhuoran29 since we added the ```heating_steam``` state block, would it be correct to just cost the ```flow_vol_phase["Vap"]``` here?

kurbansitterley: done in 045b9ab

src/watertap_contrib/reflo/unit_models/crystallizer_effect.py
```diff
@@ -0,0 +1,536 @@
...
+            iscale.set_scaling_factor(self.delta_temperature_in[0], 0.1)
+
+        if iscale.get_scaling_factor(self.delta_temperature_out[0]) is None:
+            iscale.set_scaling_factor(self.delta_temperature_out[0], 0.1)
+
+        if iscale.get_scaling_factor(self.heat_exchanger_area) is None:
+            iscale.set_scaling_factor(self.heat_exchanger_area, 0.1)
+
+        if iscale.get_scaling_factor(self.overall_heat_transfer_coefficient) is None:
+            iscale.set_scaling_factor(self.overall_heat_transfer_coefficient, 0.01)
```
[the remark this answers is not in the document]
kurbansitterley: done in 045b9ab

src/watertap_contrib/reflo/unit_models/crystallizer_effect.py
```diff
@@ -0,0 +1,536 @@
...
+
+        self.heat_exchanger_area = Var(
+            initialize=1000.0,
+            bounds=(0, None),
+            units=pyunits.m**2,
+            doc="Heat exchanger area",
+        )
+
+        self.overall_heat_transfer_coefficient = Var(
+            initialize=100.0,
```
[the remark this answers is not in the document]
kurbansitterley: done in 045b9ab

src/watertap_contrib/reflo/costing/units/multi_effect_crystallizer.py
Zhuoran29: It'd be correct. The functions in `_compute_steam_specific_energy` below is the same as in the water property package.

kurbansitterley: Ok- I believe it will change the costing results. The crystallizer prop pack doesn't have the specific volume property, but if the property model is calculating it in a different way then I suppose that is just as good. 

So in theory the ```steam_flow``` calculated here should be equal to ```flow_vol_phase["Vap"]```?

kurbansitterley: believe we figured this out.

src/watertap_contrib/reflo/costing/units/multi_effect_crystallizer.py
```diff
@@ -0,0 +1,358 @@
...
+        units=pyo.units.USD_2020 / pyo.units.kg,
+    )
+
+    costing = blk.parent_block()
+    costing.register_flow_type("NaCl_recovered", blk.cost)
+
+
+def build_steam_cost_param_block(blk):
+
+    blk.cost = pyo.Param(
```
adam-a-a, line 41: Will this show up on the model as m.fs.costing.unit_name.steam.cost?

src/watertap_contrib/reflo/costing/units/multi_effect_crystallizer.py
```diff
@@ -0,0 +1,358 @@
...
+    blk.iec_percent = pyo.Var(
+        initialize=1.43,
+        doc="Forced circulation crystallizer installed equipment cost (Diab and Gerogiorgis, 2017)",
+        units=pyo.units.dimensionless,
+    )
+
+    blk.volume_cost = pyo.Var(
+        initialize=16320,
+        doc="Forced circulation crystallizer cost per volume (Yusuf et al., 2019)",
+        units=pyo.units.USD_2007,  ## TODO: Needs confirmation, but data is from Perry apparently
```
adam-a-a, line 82: What is the todo note in reference to? The doc says Yusuf et al., 2019, is the reference, but the comment indicates otherwise. Are we talking about the value itself needing confirmation, the units (year), or something else?

kurbansitterley: I am not sure... I believe this is a direct copy/paste from the WaterTAP main crystallizer costing pkg, but I will have to check.

src/watertap_contrib/reflo/unit_models/crystallizer_effect.py
```diff
@@ -0,0 +1,430 @@
...
+            self.flowsheet().config.time,
+            doc="Material properties of pure water vapour at outlet",
+            **tmp_dict,
+        )
+
+        self.add_port(name="pure_water", block=self.properties_pure_water)
+
+        self.properties_pure_water[0].flow_mass_phase_comp["Liq", "H2O"].fix(1e-8)
+        self.properties_pure_water[0].flow_mass_phase_comp["Liq", "NaCl"].fix(0)
+        self.properties_pure_water[0].flow_mass_phase_comp["Sol", "NaCl"].fix(0)
```
adam-a-a, line 98: Is there a reason why water is 1e-8 and nacl components set to 0?

src/watertap_contrib/reflo/unit_models/crystallizer_effect.py
```diff
@@ -0,0 +1,430 @@
...
+        self.properties_pure_water[0].mass_frac_phase_comp["Liq", "NaCl"]
+        self.properties_in[0].conc_mass_phase_comp["Liq", "NaCl"]
+        self.properties_in[0].flow_vol_phase["Liq"]
+
+        self.inlet.temperature.setub(1000)
+        self.outlet.temperature.setub(1000)
+        self.solids.temperature.setub(1000)
+        self.vapor.temperature.setub(1000)
+        self.pure_water.temperature.setub(1000)
```
adam-a-a, line 107: Seems like something that should be in the documentation of this model (on the readthedocs)

src/watertap_contrib/reflo/costing/units/multi_effect_crystallizer.py
kurbansitterley: Will be ```m.fs.costing.steam.cost```

src/watertap_contrib/reflo/unit_models/crystallizer_effect.py
```diff
@@ -0,0 +1,430 @@
...
+        )
+
+        self.pump_head_height = Param(
+            initialize=1,
+            mutable=True,
+            units=pyunits.m,
+            doc="Crystallizer pump head height -  assumed, unvalidated",
+        )
+
+        self.energy_flow_superheated_vapor = Var(
```
adam-a-a, line 130: Is there more of an "IDAES naming convention" that can be used for this variable?

src/watertap_contrib/reflo/unit_models/crystallizer_effect.py
kurbansitterley: Agreed. Documentation will have to be a separate PR due to time.

src/watertap_contrib/reflo/unit_models/crystallizer_effect.py
```diff
@@ -0,0 +1,430 @@
...
+
+        @self.Constraint()
+        def eq_p_con1(b):
+            return b.pressure_operating == b.properties_out[0].pressure
+
+        @self.Constraint()
+        def eq_p_con2(b):
+            return b.pressure_operating == b.properties_solids[0].pressure
+
+        @self.Constraint()
```
adam-a-a, line 200: Is there an `eq_p_con3` this is still being used from the parent class?

kurbansitterley: Yes. Parent class has ```eq_q_con1```, ```eq_q_con2```, and ```eq_q_con3```.

src/watertap_contrib/reflo/unit_models/multi_effect_crystallizer.py
```diff
@@ -0,0 +1,335 @@
...
+                    )
+
+                @effect.Constraint(doc="Heat transfer equation for first effect")
+                def eq_heat_transfer_effect_1(b):
+                    # return b.work_mechanical[0] == pyunits.convert(
+                    #     b.overall_heat_transfer_coefficient
+                    #     * b.heat_exchanger_area
+                    #     * b.delta_temperature[0]
+                    #     # , to_units=pyunits.kilowatt
+                    # )
```
adam-a-a, line 200: remove commented code

src/watertap_contrib/reflo/unit_models/multi_effect_crystallizer.py
```diff
@@ -0,0 +1,335 @@
...
+                        )
+                    )
+
+                self.add_port(name="inlet", block=effect.properties_in)
+                self.add_port(name="outlet", block=effect.properties_out)
+                self.add_port(name="solids", block=effect.properties_solids)
+                self.add_port(name="vapor", block=effect.properties_vapor)
+                self.add_port(name="pure_water", block=effect.properties_pure_water)
+                self.add_port(name="steam", block=effect.heating_steam)
+                self.steam.temperature.setub(1000)
```
adam-a-a, line 224: Not sure I understand the logic---this looks like you add ports for only the first effect, when I would imagine the outlets being specified for the last effect. Whatever the case, can you explain the logic here?

src/watertap_contrib/reflo/unit_models/crystallizer_effect.py
kurbansitterley: This would be a question for @Zhuoran29

src/watertap_contrib/reflo/unit_models/multi_effect_crystallizer.py
```diff
@@ -0,0 +1,335 @@
...
+            optarg=optarg,
+        )
+
+        init_log = idaeslog.getInitLogger(self.name, outlvl, tag="unit")
+        solve_log = idaeslog.getSolveLogger(self.name, outlvl, tag="unit")
+
+        opt = get_solver(solver, optarg)
+        for n, eff in self.effects.items():
+            if n == 1:
+                assert degrees_of_freedom(eff.effect) == 0
```
adam-a-a, line 298: This should raise an exception if DOF do not equal 0

src/watertap_contrib/reflo/unit_models/crystallizer_effect.py
kurbansitterley: I also believe there was some discussion of this in watertap-org/watertap#1252

src/watertap_contrib/reflo/unit_models/tests/test_multi_effect_crystallizer.py
```diff
@@ -0,0 +1,1127 @@
...
+            ("pressure_sat", None): saturated_steam_pressure,
+        },
+        hold_state=True,
+    )
+    first_effect.heating_steam[0].flow_mass_phase_comp["Vap", "H2O"].unfix()
+
+    return m
+
+
+class TestMultiEffectCrystallizer_4Effects:
```
adam-a-a, line 193: So 4 effects is the default, and you test for 4 effects and 3 effects. What happens if the user tries entering 1 effect? Similarly, what about 2 effects? I went scanned through the core code quickly, so I know I missed this. I just wonder if we should just add two quick, shortened versions of these tests just to ensure that we get what we'd expect from entering 1 effect or 2 effects. For example, in 2 effects, I might expect some special handling of the first and last effect, hence interest in this case. For 1 effect, I would expect the code to handle this case or throw an exception if the multieffect model is not setup to handle specification of 1 effect.

adam-a-a (commented):
Quite an impressive bit of work here that I know must've taken a lot of thought and effort! Well done! I just have a couple of comments from a quick scan.

adam-a-a:
My other general comment is we _definitely_ want to document the single and multieffect models thoroughly--though that can happen a subsequent PR down the line and shouldn't hold this up.

src/watertap_contrib/reflo/unit_models/multi_effect_crystallizer.py
kurbansitterley: Hm. This has made me think, I will have to revisit.

src/watertap_contrib/reflo/unit_models/tests/test_multi_effect_crystallizer.py
kurbansitterley: addressed in e0b5b44.

Added test for 2 effects.
If the user tries to use ```MultiEffectCrystallizer``` with a single effect, it will raise ```ConfigurationError``` and they will be directed to use the ```CrystallizerEffect``` model.

src/watertap_contrib/reflo/unit_models/multi_effect_crystallizer.py
kurbansitterley: Ok so this comment prompted to revisit the model and realized the inlet flows the user would set would not end up being the resulting inlet flows from the model. This would obviously be problematic if this model was connected to an upstream process.

This was addressed starting in 651dba5.

I added a ```ControlVolume0D``` to ensure a mass balance across the unit. The model initialization will now automatically deactivate the constraint that links energy flows between effects and initialize the water/salt mass flow rate in each effect with ```control_volume_mass_flow / number_effects```  (regardless of if the user fixed the mass flows for individual effects in the build) as if it was a standalone model.

Then, inlet flows to each effects are unfixed, linking constraints are re-activated, and the salt _concentration_ at the inlet for each effect is fixed to be the same across all effects. When the model is solved, the ```ControlVolume0D``` handles how much flow to send to each effect. You can see different approaches to take for the build and the resulting DOF in the ```test_dof``` tests.

kurbansitterley: Oh I guess I never actually answered your question lol.

I moved the ```inlet``` and ```outlet``` to be on the control volume. As for why the rest are on the first effect, I guess I figured that is where they might be in an actual crystallizer? I don't really have a great reason. 

I actually think my moving to the control volume approach makes the ```pure_water``` port moot. I shall remove it.

src/watertap_contrib/reflo/unit_models/multi_effect_crystallizer.py
kurbansitterley: addressed

src/watertap_contrib/reflo/unit_models/tests/test_multi_effect_crystallizer.py
adam-a-a: Nice. One thought I have--say you wanted to run optimization cases across a range of conditions and a range of number of effects, including a single effect. You'd need to add some logic in to such a script to toggle between the CrystallizerEffect model for the case where a single effect is optimized, and then toggle to the MultiEffectCrystallizer model for all other cases. 

All that just to say, it might be useful to actually allow for a single effect via MultiEffectCrystallizer too. Definitely not something to consider for this PR--just food for thought.

src/watertap_contrib/reflo/unit_models/crystallizer_effect.py
```diff
@@ -0,0 +1,436 @@
...
+            units=pyunits.kilowatt / pyunits.m**2 / pyunits.degK,
+            doc="Overall heat transfer coefficient for heat exchangers",
+        )
+
+        # @self.Constraint()
+        # def eq_pure_vapor_flow_rate(b):
+        #     return (
+        #         b.properties_pure_water[0].flow_mass_phase_comp["Vap", "H2O"]
+        #         == b.properties_vapor[0].flow_mass_phase_comp["Vap", "H2O"]
+        #     )
```
adam-a-a, line 173:
```suggestion
```

src/watertap_contrib/reflo/unit_models/crystallizer_effect.py
```diff
@@ -0,0 +1,430 @@
...
+                    b.properties_pure_water[
+                        0
+                    ].dh_vap_mass_solvent  # Latent heat from the vapor
+                    + b.properties_vapor[0].enth_mass_solvent["Vap"]
+                    - b.properties_pure_water[0].enth_mass_solvent["Vap"]
+                )
+            )
+
+        self.del_component(self.eq_p_con1)
+        self.del_component(self.eq_p_con2)
```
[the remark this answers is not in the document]
adam-a-a: The main thing that changed were the units on enthalpy I think, from kJ/kg to J/kg, so increasing enthalpy bound values and initial values by 3 orders of magnitude. Scaling was adjusted in accordance as well. However, some nonobvious changes had to also be implemented to the scaling routine in order to get the model to converge and tests to pass again: https://github.com/watertap-org/watertap/pull/1462/files#diff-e2613f205b01d4e211ff0cb63a36c0e12136e96c1e13eef0dc47bf078b0b9a0aR766-R777

I am guessing you and @Zhuoran29 probably identified these changes already and still had issues with implementation, but just in case you didn't, hopefully some of that will be useful for future probing.

adam-a-a (approved):
LGTM - I had another comment or two to consider but I think this should be good to merge. We can make issues on any other gaps identified here.

src/watertap_contrib/reflo/unit_models/crystallizer_effect.py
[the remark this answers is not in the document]
kurbansitterley: Yes exactly. We identified the changes and it just wasn't cooperating. It is something we are going to have to reconcile at some point when we point to a newer watertap release but not in this PR.

src/watertap_contrib/reflo/unit_models/crystallizer_effect.py
kurbansitterley: This was modified to fix the Vap water phase in ```properties_pure_water``` to zero.

src/watertap_contrib/reflo/unit_models/tests/test_multi_effect_crystallizer.py
kurbansitterley: Thinking about this now, I don't think it would be very hard to implement, maybe a little clunky, but I do like the idea of having all the functionality for a unit model within a single model rather than directing the user to use a different unit model. A consideration for the future.

Zhuoran29:
I ran a few cases and the results are looking good to me! Also I added a test for an optimization scenario in case it's helpful to the case study. If the addition look good to you then I think it's good to go. @kurbansitterley

kurbansitterley (approved):

PR #125, opened 2024-09-10
Use black version in REFLO environment for GHA

kurbansitterley:
Edit checks.yml to use the black version specified in the environment build when doing linting test with GHA.

The modified lines in checks.yml are taken directly from checks.yml in main watertap.

### DIFF
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -19,7 +19,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: actions/setup-python@v5
         with:
-          python-version: '3.8'
+          python-version: '3.10'
       - name: Install Black
         # unlike the other jobs, we don't need to install watertap-reflo and/or all the dev dependencies,
         # but we still want to specify the Black version to use in requirements-dev.txt for local development
--- a/.github/workflows/checks.yml
+++ b/.github/workflows/checks.yml
@@ -13,7 +13,7 @@ jobs:
 
   linting:
     name: Check code is formatted (Black)
-    # OS and/or Python version don't make a difference, so we choose ubuntu and 3.8 as defaults
+    # OS and/or Python version don't make a difference, so we choose ubuntu and 3.10 as defaults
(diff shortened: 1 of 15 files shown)


.github/workflows/checks.yml
```diff
@@ -12,13 +12,24 @@ env:
 jobs:
 
   linting:
+    name: Check code is formatted (Black)
+    # OS and/or Python version don't make a difference, so we choose ubuntu and 3.8 as defaults
```
lbianchi-lbl, line 16:
```suggestion
    # OS and/or Python version don't make a difference, so we choose ubuntu and 3.10 as defaults
```

.github/workflows/checks.yml
```diff
@@ -12,13 +12,24 @@ env:
...
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v4
-      - uses: psf/black@stable
+      - uses: actions/setup-python@v5
         with:
-          options: "--check --verbose"
-          version: "~= 22.0"
+          python-version: '3.8'
```
lbianchi-lbl, line 22:
```suggestion
          python-version: '3.10'
```

.github/workflows/checks.yml
```diff
@@ -12,13 +12,24 @@ env:
...
+      - name: Install Black
+        # unlike the other jobs, we don't need to install WaterTAP and/or all the dev dependencies,
```
lbianchi-lbl, line 24:
```suggestion
        # unlike the other jobs, we don't need to install watertap-reflo and/or all the dev dependencies,
```

lbianchi-lbl (commented):
Looks good. I left some a couple of suggestion to update the Python version from 3.8 (unsupported by the latest IDAES even if in this case it doesn't technically matter), and fix a reference to WaterTAP that was left in another comment.

lbianchi-lbl (approved):

adam-a-a (approved):

PR #126, opened 2024-09-12
Point to WaterTAP 1.0.0rc0

kurbansitterley:
Temporary fix for test failures in #121 when using watertap==1.0.0 but not watertap=1.0.0rc0.


Zhuoran29 (approved):
Approve in case the merge is needed.

adam-a-a (approved):

adam-a-a:
@kurbansitterley Do we want to make sure that other REFLO PRs won't break after this dependency is introduced?

kurbansitterley:
> @kurbansitterley Do we want to make sure that other REFLO PRs won't break after this dependency is introduced?

Done.

PR #128, opened 2024-09-24
Add Deep Well Injection unit model and costing

kurbansitterley:
This PR will add a simple deep well injection unit model with three costing options:

- calculating OPEX based on user-provided LCOW (calculated such that resulting LCOW will equal this value)
- calculating CAPEX based on user-provided LCOW (calculated such that resulting LCOW will equal this value)
- costing based on BLM deep well injection costing model

TODO:

- [x] add test file
- [x] add costing option for $ per m3 disposed

### DIFF
--- a/src/watertap_contrib/reflo/costing/units/deep_well_injection.py
+++ b/src/watertap_contrib/reflo/costing/units/deep_well_injection.py
@@ -28,6 +28,7 @@ from watertap.costing.util import register_costing_parameter_block
 from ..util import (
     make_capital_cost_var,
     make_fixed_operating_cost_var,
+    make_variable_operating_cost_var,
 )
 
 __author__ = "Kurban Sitterley"
@@ -65,7 +66,8 @@ blm_costing_params_dict = {
 
 
 class DeepWellInjectionCostMethod(StrEnum):
-    simple = "simple"
+    as_capex = "as_capex"
+    as_opex = "as_opex"
     blm = "blm"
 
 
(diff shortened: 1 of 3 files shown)


src/watertap_contrib/reflo/costing/units/deep_well_injection.py
```diff
@@ -0,0 +1,458 @@
...
+    )
+
+    blk.capital_cost_constraint = Constraint(
+        expr=blk.capital_cost
+        == pyunits.convert(
+            (blk.costing_package.deep_well_injection.dwi_lcow * blk.annual_flow)
+            / blk.costing_package.capital_recovery_factor,
+            to_units=blk.costing_package.base_currency,
+        )
+    )
```
Zhuoran29, line 122: Should the `fixed_operating_cost` be included here, since `LCOW = (capex * recovery_factor + opex) / annual_flow` ?

src/watertap_contrib/reflo/costing/units/deep_well_injection.py
```diff
@@ -0,0 +1,458 @@
...
+
+    blk.fixed_operating_cost_constraint = Constraint(
+        expr=blk.fixed_operating_cost
+        == -blk.costing_package.maintenance_labor_chemical_factor * blk.capital_cost
+    )
```
Zhuoran29, line 127: Why is the `fixed_operating_cost` a negative term?

Zhuoran29 (commented):
All LGTM! Just not so sure about one of the cost equations.

src/watertap_contrib/reflo/costing/units/deep_well_injection.py
kurbansitterley: this is essentially solving the LCOW expression from ```.add_LCOW()``` method for capital cost. This is a hacky approach, the assumption being that if we are using this costing approach, we just want WaterTAP to properly aggregate the costs such that the LCOW attributable to this unit is identical to the one we set as a parameter on ```m.fs.costing.deep_well_injection.dwi_lcow```.

kurbansitterley:
Setting to draft until I can make the change we discussed.

kurbansitterley:
added ```as_opex``` costing method and ```as_capex``` costing methods.

MuktaHardikar (approved):
LGTM!

zacharybinger (approved):
I like the added flexibility

PR #131, opened 2024-10-01
Add solar still ZLD unit model, costing model, test file

kurbansitterley:
This PR adds an implementation of the ZLD solar still unit model and costing model provided by NMSU for the SETO project.

The unit model is relatively straight forward.
The most important parameter ```water_yield``` relies on user input for the weather file (.csv) and an external function in ```util``` to calculate the daily water yield. This value is used to set the model parameter ```water_yield```. 

The model assumes:
- All of the salt is removed (goes to ```properties_waste```)
- All of the water that comes in goes out as product water (i.e., 100% recovery)

The test file tests against a case provided by the NMSU team.

The regular(non-ZLD) solar still model may be a separate PR. Documentation will also be a separate PR.

TODO

- [x] verify source and collection steps for weather data to ensure it is properly formatted. The external utility function for calculating water yield relies on a specific organization of columns.


src/watertap_contrib/reflo/unit_models/tests/test_solar_still.py
```diff
@@ -0,0 +1,247 @@
...
+    inlet_dict = {
+        "solute_list": ["TDS"],
+        "mw_data": {"TDS": 31.4038218e-3},
+        "material_flow_basis": MaterialFlowBasis.mass,
+    }
+
+    water_yield_calc_dict = dict(
+        input_weather_file_path=test_data_path,  # path to input weather file
+        interval_day=15,  # interval at which to calculate water yield (e.g., every 15 days)
+        salinity=20,  # salinity of influent water; g/L
```
adam-a-a, line 66: We could probably do away with this salinity key and use total_dissolved_solids, for example, under the hood of the unit model. My broader point is whether this can be more of an optional arg that would only be specified if the salinity input couldn't be obtained from the property model by the unit model. 

This is actually a good example where you might want to manually enter the salinity as you did here since TDS is the solute, and currently, MCAS may or may not output the right value for total_dissolved_solids. (Note that I am not considering complexities such as organics contributing to TDS and just thinking of salinity here).

src/watertap_contrib/reflo/unit_models/tests/test_solar_still.py
```diff
@@ -0,0 +1,247 @@
...
+        water_depth_basin=0.02,  # depth of water in solar still basin; m
+        length_basin=0.6,  # length of each side of basin (length=width); m
```
adam-a-a, line 68: I know this naming is an attempt to follow IDAES' typical naming conventions, but since these do not correspond to any establish IDAES names, I would opt to go with more readable names, such as `basin_water_depth` and `basin_length`.

adam-a-a: Another question is whether I need to specify these design variables here or if I can set Param/Var values how I typically would on an IDAES model.

src/watertap_contrib/reflo/unit_models/tests/test_solar_still.py
```diff
@@ -0,0 +1,247 @@
...
+    rho = 1000 * pyunits.kg / pyunits.m**3
+    inlet_dict = {
+        "solute_list": ["TDS"],
+        "mw_data": {"TDS": 31.4038218e-3},
+        "material_flow_basis": MaterialFlowBasis.mass,
+    }
+
+    water_yield_calc_dict = dict(
+        input_weather_file_path=test_data_path,  # path to input weather file
+        interval_day=15,  # interval at which to calculate water yield (e.g., every 15 days)
```
adam-a-a, line 65:
```suggestion
        yield_horizon_days=15,  # interval at which to calculate water yield (e.g., every 15 days)
```

src/watertap_contrib/reflo/unit_models/tests/test_solar_still.py
```diff
@@ -0,0 +1,247 @@
...
+        assert not m.fs.unit.config.dynamic
+        assert not m.fs.unit.config.has_holdup
+        assert m.fs.unit.config.property_package is m.fs.properties
+        assert_units_consistent(m)
+
+    @pytest.mark.unit
+    def test_build(self, ss_frame):
+        m = ss_frame
+
+        port_list = ["inlet", "outlet", "waste"]
```
adam-a-a, line 122: Wondering if the port naming should follow the convention for ZO models in WaterTAP (`inlet`,`treated`,`byproduct`), unless this is really a 0D model.

src/watertap_contrib/reflo/unit_models/tests/test_solar_still.py
kurbansitterley: This has to be a manual entry since it goes into the helper function to calculate the water yield, and passing e.g., `conc_mass_phase_comp["Liq", "TDS"]` wouldn't work since the value as passed during `build` would be the initialized value of the variable rather than the actual value of TDS for the flowsheet. 

(but maybe I misunderstood your point)

src/watertap_contrib/reflo/unit_models/tests/test_solar_still.py
kurbansitterley: Pretty much same answer as before - we would be passing the initialized values in `build` to calculate water yield and if the user changes the value for each of these after the build, the calculated water yield would not reflect those different parameter values.

I suppose we could separate the two (i.e., remove the water yield calculation to a separate method that would need to be called manually after build):

```
m.fs.solar_still = SolarStill()
m.fs.solar_still.water_depth_basin.set_value(0.05)
m.fs.solar_still.water_depth_basin.length_basin(0.5)
m.fs.solar_still.calculate_water_yield()
```

src/watertap_contrib/reflo/unit_models/tests/test_solar_still.py
kurbansitterley: I have never been a fan of `treated` or `byproduct` tbh. But you are right that I don't think this is a 0D model.

PR #133, opened 2024-10-07
Add KBHDP case study component files

kurbansitterley:
This PR adds the component files for unit processes used in KBHDP case study flowsheets.


kurbansitterley:
Nothing really to review here... just want to get the component files in so that future case study PRs aren't enormous.

PR #134, opened 2024-10-09
Remove electricity_cost from REFLOSystemCosting build

kurbansitterley:
This is a small PR for a few needed bandaids: 
- remove the electricity_cost Param from REFLOSystemCosting build function that was raising errors
- fix the existing electricity_cost Var to the current value (0.0718)
- add capital_cost_constraint to DWI as_opex costing method
- change type of error raised for add_LCOE() method in REFLOSystemCosting


MuktaHardikar (approved):
Checked these changes with the FPC-MD-DWI case study and they look good!

Issue #28, opened 2023-02-20
Add test files for costing package


kurbansitterley:
Costing package doesn't currently have test files.

adam-a-a:
It looks like PR #5 added tests for the SETO costing package. Did you mean that these tests should be added to, is there another costing package that you are referring to without any testing, or can we close this issue?

kurbansitterley:
The tests added in #5 are pretty much a copy/paste from the main WaterTAP costing package. I meant it as something to build off of, but haven't returned to it. The tests there don't really touch any of the unique aspect of the SETO costing package.

PR #136, opened 2024-10-22
Updates and Improvements to REFLO System Costing

MuktaHardikar:
This PR includes changes to the REFLO costing method to track energy (electricity/heat) purchased, sold and produced/used by the treatment and energy blocks.

- Default price of heat and electricity updated to be 0
- Purchase price and sale price for heat and electricity added
- Variable to track energy purchased/sold added-Based on [Darices method](https://github.com/dguittet/watertap-seto/blob/cost_balance/src/watertap_contrib/reflo/costing/watertap_reflo_costing_package.py#L326) 
- Add initialization to REFLOSystemCosting
- Add scaling factors to REFLOSystemCosting
- Add LCOE and calculations for yearly and total electricity production
- Add LCOH and calculations for yearly and total heat production
- Add LCOT (levelized cost of treatment)
- Update LCOW
- Add many tests to test all three costing under all combinations of electricity/heat consumption/generation
- Add check for common parameter equivalence for all three costing packages in REFLOSystemCosting
- Remove ability to add_LCOW from EnergyCosting
- fix failing unit model costing tests that relied on energy prices != 0
- Add the cost of electricity and heat calculated to total operating cost

Future PR:

- Need a way to separate capital/operating costs related to electricity and heat generating systems so the proper value is used in LCOE/LCOH calculations when both are present on the energy costing block.


zacharybinger:
We should merge this asap so that way all the tests are updated for the subsequent PRs

src/watertap_contrib/reflo/costing/watertap_reflo_costing_package.py
```diff
@@ -321,11 +321,7 @@ def build_integrated_costs(self):
                 expr=(
                     self.frac_heat_from_grid
                     == 1
-                    - (
-                        -1
-                        * energy_cost.aggregate_flow_heat
-                        / treat_cost.aggregate_flow_heat
-                    )
+                    - energy_cost.aggregate_flow_heat / treat_cost.aggregate_flow_heat
```
MuktaHardikar, line 328: `energy_cost.aggregate_flow_heat` is negative. That's why it was multiplied by -1.

src/watertap_contrib/reflo/costing/solar/photovoltaic.py
```diff
@@ -92,6 +92,7 @@ def build_photovoltaic_cost_param_block(blk):
 )
 def cost_pv(blk):
 
+    blk.costing_package.has_electricity_generation = True
```
adam-a-a, line 95: Just a comment on PV costs: it would be totally fair to cost PV out on a $/kW_dc basis. I remember using numbers on the order of $1000 - $2000/kWdc

kurbansitterley: I think this is how we do it already?

kurbansitterley (approved):

kurbansitterley:
would resolve #28

zacharybinger (approved):

src/watertap_contrib/reflo/costing/watertap_reflo_costing_package.py
```diff
@@ -73,10 +118,145 @@ def build_global_params(self):
...
+
+        self.aggregate_flow_heat_sold = pyo.Var(
+            initialize=0,
+            domain=pyo.NonNegativeReals,
+            doc="Aggregated heat produced",
+            units=pyo.units.kW,
+        )
+
+        # energy producer's electricity flow is negative
+        self.aggregate_electricity_balance = pyo.Constraint(
```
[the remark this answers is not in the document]
kurbansitterley: I added a check so that a ValueError is raised if there is not energy block. Essentially REFLOSystemCosting can only be used if there are both TreatmentCosting and EnergyCosting packages available.

PR #138, opened 2024-11-05
KBHDP RPT 1

zacharybinger:
Creating PR for RPT1 Case Study

This case study includes EC, UF, RO, DWI powered by PV


src/watertap_contrib/reflo/analysis/case_studies/KBHDP/components/EC.py
```diff
@@ -40,9 +40,74 @@
...
+        print_infeasible_bounds(blk)
+        print_close_to_bounds(blk)
+        assert False
+
+
+def build_system():
+    """Function to create concrete model for individual unit model flowsheet"""
+    m = ConcreteModel()
+    m.db = REFLODatabase()
+
```
kurbansitterley, line 88: FYI I think you will need a constraint here to calculate the conductivity from the TDS, otherwise it uses the default value of 1 mS/cm. You can see what I have in the Permian component file.

kurbansitterley (approved):
Finalization of these files after we have final results.

PR #143, opened 2024-11-19
Add Evaporation Pond Model

kurbansitterley:
This PR will add the evaporation pond model and necessary updates to AWE property package to use evaporation pond model.

Evaporation pond model:
- takes in weather data, `weather` state block on model is indexed properties for all 365 days in a year
- uses BREB method to calculate evaporation rate
- adjustment factor for salinity
- adjustment factor for evaporation enhancement technology (organic dye)
- cost model based on BLM reference
- test for both enhanced and not enhanced models

AirWaterEq prop pack added:
- SaturationVaporPressureCalculation
- VaporPressureCalculation
- RelativeHumidityCalculation
- LatentHeatVaporizationCalculation
- SpecificHeatWaterCalculation
- DensityCalculation
- moved some stuff around

Changes to AirWaterEq necessitated updates to the AirStripping model and the tests for AirWaterEq and air stripping.

TODO: 

- [ ] Add documentation (in separate PR 🙃)


src/watertap_contrib/reflo/property_models/air_water_equilibrium_properties.py
```diff
@@ -919,25 +1611,87 @@ def build(self):
...
     def _dens_mass_phase(self):
 
-        self.dens_mass_phase = Var(
-            self.params.phase_list,
-            initialize=self.params.config.density_data,
-            units=pyunits.kg / pyunits.m**3,
-            doc="Mass density for each phase",
-        )
+        if self.params.config.density_calculation == DensityCalculation.calculated:
+            # TODO: add McCain relationship for higher salinity brines (>150 g/L)
```
zacharybinger, line 1615: Is this TODO completed?

kurbansitterley: No- this is something that I thought of yesterday and came across a reference for, and just didn't get in. 

Though, if you think we have need for good density calculations >150 g/L, I can get it in (I have the reference open).

zacharybinger: I think it might be good to have, but not critical for this PR. Can maybe add later, I think prioritizing merging the outstanding unit models is more important and the >150 g/L functionality can be addresses quickly later

src/watertap_contrib/reflo/unit_models/tests/test_evaporation_pond.py
```diff
@@ -0,0 +1,729 @@
...
+    def test_config(self, pond_frame):
+        m = pond_frame
+        assert len(m.fs.unit.config) == 9
+
+        assert not m.fs.unit.config.dynamic
+        assert not m.fs.unit.config.has_holdup
+        assert not m.fs.unit.config.add_enhancement
+        assert m.fs.unit.config.property_package is m.fs.properties
+        assert m.fs.unit.config.dike_height == 8
+        # BUG: smooth_min/smooth_max will always return dimensionless quantity?
```
zacharybinger, line 168: So I went down the rabbit whole on this one and figured out why the assert_units_consistent was failing. I got it to work, but that cause a bunch of the other tests to fail. So if the model matches the timeseries data then maybe we just ignore it for now?

zacharybinger: In summmary smooth_max(a,b,eps) takes that eps term which is default 1e-4 which is fine if a & b are dimensionless. But if they have units, which in this case are K, the eps has to be a Param with the same units. Thats why there assert was throwing units inconsistent.

zacharybinger (approved):
Other than the assert units thing, everything looks good

src/watertap_contrib/reflo/unit_models/tests/test_evaporation_pond.py
kurbansitterley: Hm ok, thanks for figuring that out. Let me see if I can fix it.

kurbansitterley: > Hm ok, thanks for figuring that out. Let me see if I can fix it.

Ok there are a lot of regressions used for determining parameters (e.g., `emissivity_air`) that are causing the `assert_units_consistent` to fail. And there is one that has a `smooth_min` in a `smooth_max` that could be also causing more problems... but like you said I have validated the results so I am fine with letting the test slide

PR #148, opened 2024-12-02
Add USD_2023 as currency unit to costing packages

kurbansitterley:
This PR manually adds USD_2023 as a unit to all costing packages using the load_definition_from_strings method for pint. The existing costing package only has USD up to 2022. 

The definition is taken directly from idaes > core > base > costing_base.py.


src/watertap_contrib/reflo/costing/tests/test_reflo_watertap_costing_package.py
```diff
@@ -332,7 +332,7 @@ def build_heat_and_elec_gen():
     #### TREATMENT BLOCK
     m.fs.treatment = Block()
     m.fs.treatment.costing = TreatmentCosting()
-    m.fs.treatment.costing.base_currency = pyunits.USD_2002
+    m.fs.treatment.costing.base_currency = pyunits.USD_2023
```
adam-a-a, line 335: Shouldn't this be set as default in the costing package and propagate down to the two different Costing blocks?

src/watertap_contrib/reflo/costing/watertap_reflo_costing_package.py
```diff
@@ -377,6 +379,8 @@ class REFLOSystemCostingData(WaterTAPCostingBlockData):
     def build_global_params(self):
         super().build_global_params()
 
+        pyo.units.load_definitions_from_strings(["USD_2023 = 500/797.9 * USD_CE500"])
+
         self.base_currency = pyo.units.USD_2021
```
adam-a-a, line 384: Shou
```suggestion
        self.base_currency = pyo.units.USD_2023
```
Are you leaving this in purposely to avoid breaking tests with a new default?

kurbansitterley: Yes exactly. Setting 2023 as the default would cause many unit model costing tests to fail and I just don't want to deal with that at the moment.

kurbansitterley: I suppose an alternative would be to put `m.fs.costing.base_currency = pyunits.USD_2021` in all those tests

adam-a-a (commented):
I would assume that setting the base currency for SystemCosting should propagate down to Treatment and Energy Costing classes. Or, setting base currency for one should set for the others.

src/watertap_contrib/reflo/costing/watertap_reflo_costing_package.py
adam-a-a: That would be an alternative. Then you could add a test simply to verify the default currency for all 3 Costing blocks

adam-a-a:
> I would assume that setting the base currency for SystemCosting should propagate down to Treatment and Energy Costing classes. Or, setting base currency for one should set for the others.

Probably not for this PR, but something to consider for the long run. Allowing the user to choose base currency for one leaves open room for error, e.g., when the user forgets to set the same base currency for other costing blocks. Ideally, you should be able to set one global base currency and call it a day.

kurbansitterley:
> I would assume that setting the base currency for SystemCosting should propagate down to Treatment and Energy Costing classes. Or, setting base currency for one should set for the others.

I agree that would be ideal. But, the issue with this is that `REFLOSystemCosting` build relies on grabbing the treatment and energy costing blocks. So those blocks must be created to use `REFLOSystemCosting` and thus setting `base_currency` post build of `REFLOSystemCosting` would not propagate to the `TreatmentCosting` and `EnergyCosting` blocks. The implementation I came up with is for `REFLOSystemCosting` to first assert that `base_currency` (and some other parameters) are equivalent for `TreatmentCosting` and `EnergyCosting` and then to set the `REFLOSystemCosting` parameter to be equivalent. I am open to different implementations, however.

src/watertap_contrib/reflo/costing/watertap_reflo_costing_package.py
kurbansitterley: see the `_check_commpon_param_equivalence` method on `REFLOSystemCosting`

kurbansitterley: see the `_check_common_param_equivalence` method for `REFLOSystemCosting`

kurbansitterley:
> > I would assume that setting the base currency for SystemCosting should propagate down to Treatment and Energy Costing classes. Or, setting base currency for one should set for the others.
> 
> My quoted comment above is probably not for this PR, but something to consider for the long run. Allowing the user to choose base currency for one leaves open room for error, e.g., when the user forgets to set the same base currency for other costing blocks. Ideally, you should be able to set one global base currency and call it a day.

Maybe the move is to just make a new method like `set_global_parameter` or something on `REFLOSystemCosting` that would essentially do this rather than relying on the user to set these parameter values prior to building `REFLOSystemCosting`...

adam-a-a:
> The implementation I came up with is for REFLOSystemCosting to first assert that base_currency (and some other parameters) are equivalent for TreatmentCosting and EnergyCosting and then to set the REFLOSystemCosting parameter to be equivalent. I am open to different implementations, however.

Actually, that works too, assuming that system costing will raise an exception if base_currency is different between treatment and energy costing. However, is there ever a case when you might only use treatment or energy costing alone, i.e., without system costing?

kurbansitterley:
> > The implementation I came up with is for REFLOSystemCosting to first assert that base_currency (and some other parameters) are equivalent for TreatmentCosting and EnergyCosting and then to set the REFLOSystemCosting parameter to be equivalent. I am open to different implementations, however.
> 
> Actually, that works too, assuming that system costing will raise an exception if base_currency is different between treatment and energy costing. However, is there ever a case when you might only use treatment or energy costing alone, i.e., without system costing?

The first thing that the `_check_common_param_equivalence` does is check if the param is equivalent between the treatment and energy costing block. I initially had it raise an exception if this was not the case but this was changed to a logged warning since there _could_ be a scenario where we would want different energy prices on each for calculating levelized costs for each system in isolation (e.g., calculating LCOW on treatment costing with a non-zero energy price for comparison to the aggregated LCOW calculating on system costing). This is not ideal, but ultimately we decided to rely on the user to set the proper parameter values, and since there are only 4-5 users at the moment that are all in communication, we felt this was an acceptable temporary solution. Additionally, ultimately the reported TEA parameters will _likely_ come from the system costing block, and those are all assiduously converted to `base_currency` for the system costing block, so ensuring all three have the same global parameters is not too terribly critical (though I do agree there should be a more rigorous check implemented). (And looking at this implementation now, I can see that I created the `warning_msg` but only added the log line for the first check). The PR that added all this was huge and, admittedly, not thoroughly reviewed 🙃 so please point out issues or other implementations.

For your second question - You would use `TreatmentCosting` and/or `EnergyCosting` alone in an identical fashion as the `WaterTAPCosting` package. All of the unit model tests use one or the other in isolation with no issues.

src/watertap_contrib/reflo/costing/tests/test_reflo_watertap_costing_package.py
kurbansitterley: Done

kurbansitterley:
@adam-a-a anything else for this one? I'd like to get it in for the case studies

PR #161, opened 2024-12-30
SOA v3

zacharybinger:
A few more updates for the KBHDP related files


kurbansitterley:
We will get this in with the case study PRs.

PR #162, opened 2025-01-07
Update gitignore

kurbansitterley:
Add csv, png, pkl, rtf to gitignore


lbianchi-lbl:
Changing `ubuntu-latest` with `ubuntu-22.04` should resolve the Pynumero error:

https://github.com/watertap-org/watertap-reflo/blob/a52a1c94563869b995557cc53be89b5589147f7e/.github/workflows/checks.yml#L54

kurbansitterley:
moot because it was done in #163

PR #163, opened 2025-01-07
Multi Effect Crystallizer Constraint Move & Option to Cost Work as Heat or Steam

kurbansitterley:
This PR moves some of the constraints that were placed on each effect to the main unit model (i.e., `self.constr` rathter than `self.effects[n].constr` and updates the tests to reflect this change. 

This modification does not cause in any unit model performance or costing results to change.

This PR also adds an option to the multi-effect crystallizer costing model to cost the heat required as either a "steam" flow or a "heat" flow. This feature was added so the multi-effect crystallizer model can be integrated with solar models that use `aggregate_flow_heat` (not steam) to determine the heat requirement from the solar model.

Also adds a test to test this costing approach.


kurbansitterley:
#165 merged into this PR.

Zhuoran29:
LGTM! Thanks for the adaptions.

PR #165, opened 2025-01-27
Multi-effect Crystallizer Cost Work as Steam or Heat

kurbansitterley:
This PR adds an option to the multi-effect crystallizer costing model to cost the heat required as either a "steam" flow or a "heat" flow. This feature was added so the multi-effect crystallizer model can be integrated with solar models that use `aggregate_flow_heat` (not steam) to determine the heat requirement from the solar model.

Also adds a test to test this costing approach.


Zhuoran29 (approved):
LGTM!

kurbansitterley:
Changes from this PR moved to #163.

PR #167, opened 2025-01-28
Updates needed to use watertap==1.2.0 as dependency

kurbansitterley:
This PR updates units in some variables and constraints in both the `CrystallizerEffect` and `MultiEffectCrystallizer` model necessary to use `watertap==1.2.0` as dependency for WaterTAP-REFLO. For this reason, this PR also stops testing against python=3.8 since there is no watertap release for 3.8. This PR also points to the idaes and pyomo releases currently (Jan 2025) used in watertap.

There were some unit changes made in the crystallizer property package before the watertap 1.0.0 release that were not compatible with the first implementation of both of these crystallizer models. At the time, we decided to just use 1.0.0rc0 and punted making the changes necessary for using the 1.0.0 release. This PR makes those changes necessary.

Primary motivation for using watertap>=1.0.0 is inclusion of the LCOW breakdowns that were implemented in the WaterTAP costing package in 1.0.0, which will be inherited by the REFLO costing packages once this PR is merged.


src/watertap_contrib/reflo/unit_models/crystallizer_effect.py
```diff
@@ -159,9 +159,9 @@ def build(self):
         )
 
         self.overall_heat_transfer_coefficient = Var(
-            initialize=0.1,
+            initialize=1e3,
             bounds=(0, None),
-            units=pyunits.kilowatt / pyunits.m**2 / pyunits.degK,
+            units=pyunits.watt / pyunits.m**2 / pyunits.degK,
```
adam-a-a, line 164: Good catch-- I remember the only thing that was really changed previously that would have impacted this issue was the crystallization property model, where units were changed from kW to W. Thus, I am assuming that this was one of the main issues (which I figured could have been caught by an assert_units_consistent).

adam-a-a (approved):
LGTM

zacharybinger (approved):

PR #183, opened 2025-09-10
Fix Alkalinity Waste Mass Flow for Chemical Softening

kurbansitterley:
If the calculated effluent alkalinity mass flow is higher than the inlet mass flow, the waste alkalinity mass flow will be negative, resulting in infeasible solve. 

This PR uses `smooth_max` to ensure the waste alkalinity mass flow is always positive. If `IN - OUT < 0.01 * IN`, the waste flow will be `0.01 * IN`.

A possible different approach to this would be to calculate the inlet concentration/mass flow of alkalinity in the unit model itself with the same approach that the effluent is being calculated, but this could cause confusion for the user as they would have to leave the inlet alkalinity mass flow unfixed. This could also cause issues with initialization.


kurbansitterley:
Actually this is only an issue if you set `ca_eff_target` and/or `mg_eff_target` to be too high.

